PR #6, opened 2016-07-26
rjson value as string

whd:
It might be useful to pull out a subset of the JSON as a string similar to what we do here: https://github.com/mozilla-services/data-pipeline/blob/master/heka/sandbox/decoders/extract_telemetry_dimensions.lua#L212-L218.

### DIFF
--- a/rjson/index.md
+++ b/rjson/index.md
@@ -4,7 +4,7 @@
 Lua wrapper for the RapidJSON library allowing for for JSON-Schema validation
 and more efficient manipulation of large JSON structures where only a small
 amount of the data is consumed within the Lua script.  Schema pattern matching
-is restricted to a subset of regex decribed in the *Regular Expression* section
+is restricted to a subset of regex described in the *Regular Expression* section
 of the [RapidJSON Schema Documentation](http://rapidjson.org/md_doc_schema.html).
 
 ## Module
@@ -163,7 +163,7 @@ assert("bar" == str, tostring(str))
   inner expression) e.g., str = doc:value(doc:find("foo")) or "my default"
 
 *Return*
-* primitive - string, number, bool, nil or throws an error if not convertable
+* primitive - string, number, bool, nil or throws an error if not convertible
   (object, array)
 
 #### type
@@ -186,7 +186,7 @@ assert(t == "object", t)
 
 #### iter
 
-Retrieves an interator function for an object/array.
+Retrieves an iterator function for an object/array.
 
 ```lua
 local v = doc:find("obj", "arr")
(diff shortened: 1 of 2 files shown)


trink:
The API already implements a more efficient version of this (the conversion is done on the C side during output so it doesn't have copy the string back to Lua).  The telemetry landfill sandbox should be doing it as I was going to point to it as an example but it leaves the remainder in doc and never updates the submission field.  If we want it to work like the old version above it should remove the submission field and write the remainder to the Payload header (I will have a PR for that tomorrow morning).

The SHA dbfd929 portion of this PR is desirable though

whd:
After poking around a bit the problem seems to indeed be exactly with compatibility. I was looking at the telemetry landfill sandbox and followed the code there but it wasn't working for me... which is why I prepared this PR. Looking deeper, it seems that this is because if `msg.Payload` is set to user data it's not treated the same way as (say) `msg.Fields[Payload]`, where the latter is actually encoded correctly by finding the appropriate output function with `lsb_get_output_function`. Encoding of `msg.Payload` just gets dropped because the underlying lua type for the output of `doc:remove()` is not a string.

So what I want is for `msg.Payload = <userdata>` to encode the data the same way as `msg.Fields.Payload = <userdata>`. This came up because I was trying to avoid having to rewrite a different filter that was relying on `msg.Payload`. I can change the filter if we want to move it to `msg.Fields[submission]` but this is definitely some kind of backwards compatibility issue. If the assignment of `msg.Payload` must be a lua string, there is no way that I can see to use the "efficient" output function there.

trink:
Yeah, the problem with that is userdata encoding is treated as a binary string and msg.Payload is considered a UTF-8 string.  If it is not actually UTF-8 the python protobuf decoder will throw an error (which is why we moved the original gzip data from msg.Payload to msg.Fields[submission]).

whd:
I guess I will have to change the filter then to work around this particular incompatibility. Hopefully it was the only filter looking at `Payload`.

trink:
The non sandbox version could benefit from a __tostring metamethod.  That is we can rename it and keep your work if you are interested.

trink:
Final thoughts: 
- The best long term solution for the telemetry data is to put the 'remainder' into msg.Fields[submission]
- The value_as_string functionality in a non Lua sandbox is interesting but the the existing value() function should do the work and just return the serialized JSON string instead of throwing an error on an object/array.  **However**, I vote for not adding code we don't have a use for at the moment
- The PR should be modified (or a new one opened and this one closed) to include the doc typo fixes

PR #8, opened 2016-07-27
new infra updates

whd:
WIP

### DIFF
--- a/moz_telemetry/modules/moz_telemetry/extract_dimensions.lua
+++ b/moz_telemetry/modules/moz_telemetry/extract_dimensions.lua
@@ -1,0 +1,389 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Mozilla Telemetry Decoder Module
+
+## decoder_cfg Table
+```lua
+-- String used to specify the schema location on disk.
+schema_path = "/mnt/work/schemas"
+
+-- String used to specify the message field containing the user submitted telemetry ping.
+content_field = "Fields[content]" -- optional, default shown
+
+-- String used to specify the message field containing the URI of the submitted telemetry ping.
+uri_field = "Fields[uri]" -- optional, default shown
+
+-- String used to specify GeoIP city database location on disk.
+city_db_file = "/mnt/work/geoip/city.db" -- optional, if not specified no geoip lookup is performed
+
+```
+
+## Functions
+
+### transform_message
+
+Transform and inject the message using the provided stream reader.
+
+*Arguments*
+- hsr (hsr) - stream reader with the message to process
+
+*Return*
+- none, injects an error message on decode failure
+
+### decode
+
+Decode and inject the message given as argument, using a module-internal stream reader.
+
+*Arguments*
+- msg (string) - binary message to decode
+
+*Return*
+- none, injects an error message on decode failure
+--]]
+
+-- Imports
+local rjson  = require "rjson"
+local io     = require "io"
+local lpeg   = require "lpeg"
+local string = require "string"
+local table  = require "table"
+local os     = require "os"
+local floor  = require "math".floor
+local crc32  = require "zlib".crc32()
+local mtn    = require "moz_telemetry.normalize"
+local dt     = require "lpeg.date_time"
+
+local read_config          = read_config
+local assert               = assert
+local pairs                = pairs
+local ipairs               = ipairs
+local create_stream_reader = create_stream_reader
+local inject_message       = inject_message
+local type                 = type
+local tostring             = tostring
+local pcall                = pcall
+local geoip
+local city_db
+
+-- create before the environment is locked down since it conditionally includes a module
+local function load_decoder_cfg()
+    local cfg = read_config("decoder_cfg")
+    assert(type(cfg) == "table", "decoder_cfg must be a table")
+    assert(type(cfg.schema_path) == "string", "schema_path must be set")
+
+    -- the old values for these were Fields[submission] and Fields[Path]
+    if not cfg.content_field then cfg.content_field = "Fields[content]" end
+    if not cfg.uri_field then cfg.uri_field = "Fields[uri]" end
+
+    if city_db_file then 
+        geoip = require "geoip.city"
+        city_db = assert(geoip.open(cfg.city_db_file))
+    end
+
+    return cfg
+end
+
+
+local M = {}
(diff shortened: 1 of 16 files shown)


whd:
The extract_dimensions module is supposed to work with both old and new landfill types, but I've only tested it with new. The geoip config should probably be optional since if we're processing old style landfill we won't need to do the lookups.

moz_telemetry/modules/moz_telemetry/extract_dimensions.lua
```diff
@@ -0,0 +1,355 @@
...
+local mtn    = require "moz_telemetry.normalize"
+local dt     = require "lpeg.date_time"
+local geoip  = require "geoip.city"
+
+local read_config          = read_config
+local assert               = assert
+local pairs                = pairs
+local ipairs               = ipairs
+local create_stream_reader = create_stream_reader
+local inject_message	   = inject_message
```
trink: alignment

trink:
bump the package version to 1.0.2

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_payload_size.lua
```diff
@@ -0,0 +1,46 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
```
trink: Needs an H1 entry

moz_telemetry/modules/moz_telemetry/extract_dimensions.lua
```diff
@@ -0,0 +1,355 @@
...
+            local ok, err = pcall(inject_message, msg)
+            if not ok then
+                emsg.Fields.DecodeErrorType = "inject_message"
+                emsg.Fields.DecodeError = err
+                pcall(inject_message, emsg)
+            end
+        end
+    end
+
+    return 0
```
trink: Don't need the return value

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_payload_size.lua
```diff
@@ -0,0 +1,46 @@
...
+        submissionDate = "",
+    }
+}
+
+function process_message()
+    msg.Timestamp = read_message("Timestamp")
+    msg.Fields.build = read_message("Fields[appBuildId]")
+    msg.Fields.channel = read_message("Fields[appUpdateChannel]")
+    msg.Fields.docType = read_message("Fields[docType]")
+    msg.Fields.size = read_message("Fields[Size]")
```
trink: Can use read_message("size").

moz_telemetry/sandboxes/heka/input/moz_telemetry_heka_kafka.lua
```diff
@@ -0,0 +1,73 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
trink: This seems like the kafka/sandboxes/heka/input/heka_kafka.lua should be generalized to accepted a decoder and if nil uses the existing behavior (pass through injection).  To keep the terminology the same as Heka it should be a 'decoder' module.

moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
```diff
@@ -69,6 +75,11 @@ if compression and compression ~= "zst" and compression ~= "gz" then
     error("compression must be nil, zst or gz")
 end
 
+local storage_class         = read_config("storage_class") or "STANDARD"
+if storage_class and storage_class ~= "STANDARD" and
```
trink: don't need the extra storage_class check here

moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
```diff
@@ -153,6 +164,8 @@ local function get_entry(path)
 end
 
 local dimensions = mts3.validate_dimensions(read_config("dimension_file"))
+-- create the batch directory if it does not exist
+os.execute(string.format("mkdir -p %s", batch_dir))
```
trink: probably want to test the return code

moz_telemetry/sandboxes/heka/output/moz_telemetry_s3_webrtc.lua
```diff
@@ -0,0 +1,240 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
trink: duplicates too much code with moz_telemetry_s3

moz_telemetry/modules/moz_telemetry/webrtc.lua
```diff
@@ -0,0 +1,60 @@
...
+    if type(payload) ~= "table" then return false end
+    local w = payload["webrtc"] or {}
+    local i = w["IceCandidatesStats"] or {}
+    if next(i["webrtc"] or {}) or next(i["loop"] or {}) then
+        return true
+    end
+    return false
+end
+
+function filter_message ()
```
trink: Naming is a bit counter intuitive.  i.e. `if filter_message()`  to me true implies that the message should be filtered out (not kept)

moz_telemetry/modules/moz_telemetry/extract_dimensions.lua
```diff
@@ -60,7 +60,11 @@ local schema_path   = read_config("schema_path") or error("schema_path must be s
 local content_field = read_config("content_field") or "Fields[content]"
 local uri_field = read_config("content_field") or "Fields[uri]"
 
-local city_db = assert(geoip.open(read_config("geoip_city_db")))
+local city_db_file = read_config("geoip_city_db")
```
trink: You can actually make the inclusion of the entire module optional.  i.e. if you aren't using geoip decoding you don't even have to have load the `.so`

trink:
Lets formally document the splitter?, filter?, decoder, and encoder APIs recommendations (I don't want you to have to modify this code again).   Currently we have something basic like this:

```lua
-- xxx_decoder_module.lua
function decode(string) -- turns the string into something consumable by inject_message
  -- transform string
  if not filter() then
inject_message 
  return nil
end

-- xxx_encoder_module.lua
function encode() -- no parameters as it works on the current message
  -- transform message
  return string or nil (filtered) or throw error
end
```

Splitters: we haven't addressed record separation at all and it will be very interesting in Lua (a good starting place may be to look at  http://w3.impa.br/~diego/software/luasocket/ltn12.html)

Filtering: in many cases it could just be part of the decoder/encoder but in some cases you will want it to run before (like the webrtc output)

@rafrombrc

PR #9, opened 2016-08-04
Include the full raw message in the error message

mreid-moz:
It's easier to investigate / debug if you can look at the original
message that caused an error.


mreid-moz:
Closed in favour of #13

PR #15, opened 2016-08-08
Cleanup for moz sandbox testing

trink:
- no extract dimensions module on pass thru
- just use the address for the Host header


mreid-moz:
lgtm

PR #21, opened 2016-08-26
Ensure the batch file is emptied on shutdown

bbinet:
When the sandbox is being shutdown and the `flush_on_shutdown` option is set, the batch file need to be emptied or its content will be sent twice to elastic search.


trink:
Please bump the version to 1.0.2

bbinet:
Done

PR #39, opened 2016-10-03
Parity with current infra decoding

whd:
This takes a bunch of logic from https://github.com/mozilla-services/data-pipeline/blob/master/heka/sandbox/decoders/extract_telemetry_dimensions.lua to make sure the new infra output resembles the current output as closely as possible.

It also makes sure we don't throw away known ping types that don't have a schema. This can probably be improved to take as config a mapping of ping types to schemas so that it's not all hard-coded here.

@trink r?

### DIFF
--- a/moz_telemetry/CMakeLists.txt
+++ b/moz_telemetry/CMakeLists.txt
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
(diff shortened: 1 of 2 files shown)


trink:
That file has been moved in the luasandbox_1_2 branch https://github.com/mozilla-services/lua_sandbox_extensions/compare/luasandbox_1_2?expand=1#diff-ca189279af5f2a790969e2f42e75445f.  
Do you want to make a PR against that branch or do you me to apply the changes.

PR #41, opened 2016-10-04
Vacuous schema

whd:
Same as #39, against a different branch.


trink:
Looks goods, I also tested it with the latest packages and the pending lua_sandbox 1.2 version

PR #42, opened 2016-10-05
Fix geoip and crc32 logic

whd:


whd:
Also from https://github.com/brimworks/lua-zlib it looks like we need to recreate the crc32 every sample to prevent it from treating the clientids as a stream.

PR #49, opened 2016-11-23
Extract payload objects from saved-session pings

whd:
Per https://github.com/mozilla-services/data-pipeline/blob/master/heka/sandbox/decoders/extract_telemetry_dimensions.lua#L206


trink:
Can you make the PR against the sprint_nov14 branch and then it will be rolled into 1.2.2 moz_telemetry package today.

whd:
Done.

PR #67, opened 2017-01-13
Alternative fix for #65

### LINKED ISSUE #65: heka_tcp output plugin terminates if server is not listening anymore
Today I noticed that if I temporarily stop my heka_tcp input plugin, then the heka_tcp ouput plugin that were sending data to that stopped plugin will terminate quickly with the following error in logs:
```
1484245022748499356 [error] output.heka_tcp terminated: process_message() instruction_limit exceeded
```

Note that my heka_tcp input/output plugin are configured with ssl support.

bbinet:
Closes #65

### DIFF
--- a/socket/sandboxes/heka/output/heka_tcp.lua
+++ b/socket/sandboxes/heka/output/heka_tcp.lua
@@ -59,16 +59,16 @@ end
 
 local client, err = create_client()
 
-local function send_message(msg, i)
-    local len, err, i = client:send(msg, i)
-    if not len then
-        if err == "timeout" or err == "closed" then
+local function send_message(msg)
+    local i, len, err = 0
+    repeat
+        len, err, i = client:send(msg, i + 1)
+        if not len then
             client:close()
             client = nil
             return -3, err
         end
-        return send_message(msg, i)
-    end
+    until len == #msg
     return 0
 end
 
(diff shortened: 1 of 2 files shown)


socket/sandboxes/heka/output/heka_tcp.lua
```diff
@@ -59,16 +59,16 @@ end
...
 
-local function send_message(msg, i)
-    local len, err, i = client:send(msg, i)
-    if not len then
-        if err == "timeout" or err == "closed" then
+local function send_message(msg)
+    local i, len, err = 0
+    repeat
+        len, err, i = client:send(msg, i + 1)
+        if not len then
```
trink: Looking at https://github.com/diegonehab/luasocket/blob/master/src/buffer.c#L75 any error (even a recoverable one) would have a nil len so we would never retry any partial result

bbinet: Ok, what about partial success?

trink: I haven't seen a  partial send in testing or our production tests yesterday (10s of million of messages with a decent size variance) so if it even happens at all it is rare and we will just treat it as a fatal error and let HS deal with the retry.  As for the receiving end any corruption in the stream (partial message) will be logged and the data discarded.

bbinet: Great, thanks.

PR #76, opened 2017-02-17
add a heka message decoder config that doesn't clobber the message metadata

relud:
There is useful metadata about a message that is stored in default_headers, which is clobbered by the default heka json decoder. For the cloudops' logging pipeline it's important to preserve those fields, because they are required in order to properly route messages.


relud:
[access.lua](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/lpeg/io_modules/decoders/nginx/access.lua) preserves everything except timestamp. I think that's reasonable, but I'm not sure if it's clear whether this should happen when setting a config `preserve_metadata`

relud:
I'm not sure, so I'll make it configurable

relud:

PR #77, opened 2017-02-17
add support for outputting msg.Fields from s3_parquet.lua

relud:
intended for use in conjunction with #76


relud:

PR #78, opened 2017-02-17
Optionally format s3_path_dimensions using os.date

relud:


relud:

PR #79, opened 2017-02-17
add a heka message decoder config that doesn't clobber the message metadata

relud:
There is useful metadata about a message that is stored in default_headers, which is clobbered by the default heka json decoder. For the cloudops' logging pipeline it's important to preserve those fields, because they are required in order to properly route messages.

### DIFF
--- a/heka/io_modules/decoders/heka/json.lua
+++ b/heka/io_modules/decoders/heka/json.lua
@@ -73,7 +73,7 @@ function decode(data, dh)
             for k,v in pairs(msg.Fields) do
                 msg[fields_prefix..k] = v
             end
-            msg.Fields = none
+            msg.Fields = nil
         end
         msg = {Fields=msg}
         if cfg.preserve_metadata_use_timestamp then
--- a/heka/io_modules/decoders/heka/json.lua
+++ b/heka/io_modules/decoders/heka/json.lua
@@ -70,7 +70,7 @@ function decode(data, dh)
     local msg = cjson.decode(data)
     if cfg.preserve_metadata then
         if type(msg.Fields) == "table" then
-            for k,v in pairs(msg.Fields.Fields) do
+            for k,v in pairs(msg.Fields) do
                 msg[fields_prefix..k] = v
             end
             msg.Fields = none
--- a/heka/io_modules/decoders/heka/json.lua
+++ b/heka/io_modules/decoders/heka/json.lua
@@ -12,7 +12,21 @@ to inject_message so it needs to decode into a Heka message table described
 here: https://mozilla-services.github.io/lua_sandbox/heka/message.html
 
 ## Decoder Configuration Table
-* none
+
+```lua
+decoders_heka_json = {
+  -- Preserve the default_headers passed to decode by storing the json message
+  -- in Fields, after flattening the json message with a delimiter.
+  preserve_metadata = false, -- default
+
+  -- Use the Timestamp from json when preserve_metadata is true.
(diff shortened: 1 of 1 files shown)


trink:
This decoder is for processing messages that conform to the Heka JSON schema.  It appears this is some derivation of that schema and should really be its own cloudops decoder.

relud:
I'm fine making a separate decoder, but the schema of the logs is the same, the difference is where I put it in the heka message, in order to preserve `default_headers`, which I would like a way to preserve. Perhaps allowing a prefix under with it would store the `default_headers` as metadata, like `metadata_group` in [s3_parquet.lua#L60](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/parquet/sandboxes/heka/output/s3_parquet.lua#L60)

relud:
It would be nice to be able to have access to `default_headers` for `Type` `Logger` and `Timestamp` as well as the values for those that are in the log message itself, and potentially access either or both from `message_matcher`

heka/io_modules/decoders/heka/json.lua
```diff
@@ -34,18 +48,38 @@ Decode and inject the resulting message
...
+fields_prefix = "Fields" .. cfg.flatten_delimiter
+
 local M = {}
 setfenv(1, M) -- Remove external access to contain everything in the module
 
 function decode(data, dh)
     local msg = cjson.decode(data)
+    if cfg.preserve_metadata then
+        if type(msg.Fields) == "table" then
+            for k,v in pairs(msg.Fields.Fields) do
```
trink: I need to see your original input as this implies it does not conform to the Heka message schema as there is no nested Fields of Fields hash.

trink:
> It would be nice to be able to have access to default_headers
I don't understand what this comment means as you have access to everything in the current message and the current sandbox environment.  And as far as the message matcher goes this is an input decoder, Anything that is put into the newly created message is easily accessible through the message matcher downstream; if it is well structured (not a sub encoding like a JSON string in the Heka structure and even then it is still accessible but only from a string match level)

heka/io_modules/decoders/heka/json.lua
relud: That is a typo, it is supposed to be msg.Fields

heka/io_modules/decoders/heka/json.lua
```diff
@@ -34,18 +48,38 @@ Decode and inject the resulting message
...
+cfg.flatten_delimiter = cfg.flatten_delimiter or "."
+assert(type(cfg.flatten_delimiter) == "string", module_cfg .. ".flatten_delimiter must be a string")
+fields_prefix = "Fields" .. cfg.flatten_delimiter
+
 local M = {}
 setfenv(1, M) -- Remove external access to contain everything in the module
 
 function decode(data, dh)
     local msg = cjson.decode(data)
+    if cfg.preserve_metadata then
```
relud: > I don't understand what this comment means as you have access to everything in the current message and the current sandbox environment.

`dh.Type` and `msg.Type` may have different values here.

With the previous behavior there was no way to preserve `dh.Type` so that it could be used in a message matcher on an output.

With the new behavior proposed here, setting `cfg.preserve_metadata` would allow them to be referenced in a message matcher as `Type` and `Fields[Type]` respectively, with the caveat that `msg.Fields.agent` (or similar) would have to be referenced in a message matcher as `Fields[Fields.agent]`

relud:
closing this in favor of a new decoder

PR #80, opened 2017-02-17
add support for outputting msg.Fields from s3_parquet.lua

relud:
intended for use in conjunction with #79

### DIFF
--- a/parquet/sandboxes/heka/output/s3_parquet.lua
+++ b/parquet/sandboxes/heka/output/s3_parquet.lua
@@ -309,14 +309,17 @@ local function load_msg_fields()
     return record
 end
 
+local load_record
+if json_objects then
+    load_record = load_json_objects
+elseif heka_msg_use_fields then
+    load_record = load_msg_fields
+end
+
 function process_message()
     local ok, err, record
-    if json_objects or heka_msg_use_fields then
-        if json_objects then
-            record, err = load_json_objects()
-        elseif heka_msg_use_fields then
-            record, err = load_msg_fields()
-        end
+    if load_record then
+        record, err = load_record()
         if err then return -1, err end
         if load_metadata then
             record[metadata_group] = load_metadata()
@@ -327,7 +330,7 @@ function process_message()
     local writer = get_writer(path)
     local w = writer[1]
 
-    if json_objects or heka_msg_use_fields then
+    if load_record then
         ok, err = pcall(w.dissect_record, w, record)
     else
         ok, err = pcall(w.dissect_message, w)
--- a/parquet/sandboxes/heka/output/s3_parquet.lua
+++ b/parquet/sandboxes/heka/output/s3_parquet.lua
@@ -67,6 +67,10 @@ metadata_group = "metadata"
 -- If not specified the schema is applied directly to the Heka message.
 json_objects = {"Fields[submission]", "Fields[environment.system]"}
 
+-- When json_objects is not specified, apply the schema to Fields instead of
+-- the entire heka message.
+heka_message_use_fields = true -- default false
+
 s3_path_dimensions  = {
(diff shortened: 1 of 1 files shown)


parquet/sandboxes/heka/output/s3_parquet.lua
```diff
@@ -126,7 +130,8 @@ if type(json_objects) == "table" then
     require "cjson"
     json_objects_len = #json_objects
 end
-if not json_objects and metadata_group then error("metadata_group cannot be configured without json_objects") end
+local heka_msg_use_fields   = read_config("heka_message_use_fields")
+if not json_objects and not heka_msg_use_fields and metadata_group then error("metadata_group cannot be configured without json_objects") end
```
trink: This reads a bit weird.. you will need different checks as this one allows both json_objects and heka_msg_use_fields
and a metadata/heka_msg_use_fields combo

parquet/sandboxes/heka/output/s3_parquet.lua
```diff
@@ -283,10 +288,38 @@ local function load_json_objects()
     return root, nil
 end
 
+local function load_msg_fields()
+    local raw = decode_message(read_message("raw")).Fields
```
trink: fields would be a better name

relud: good point, will fix

parquet/sandboxes/heka/output/s3_parquet.lua
relud: i'll make them separate checks

trink:
I am still unclear why you don't use the Heka message schema is it simply to avoid a column name of Fields.Foo and just have a column called Foo?  If so I really don't think it is work taking a performance hit of 5-10x

relud:
It's about preserving message metadata. #79 puts everything in `Fields` to preserve metadata, and this pulls everything out of `Fields`, to get the same behavior

trink:
What does the parquet schema look like are you putting it all in a map?

relud:
right now schemas we are using look like this: https://github.com/mozilla/normandy/blob/master/recipe-server/parquet_schemas/normandy.app.docker.app/request.summary.cfg#L12-L39

relud:
but that requires code not in this PR, like this:
<pre>
    local raw_fields = decode_message(read_message("raw")).Fields
    -- if the message didn't have Fields then we can't we can't encode the message, skip it
    if not raw_fields then return nil end

    -- convert from raw message table fields to a hash http://mozilla-services.github.io/lua_sandbox/heka/message.html#array-based-message-fields
    msg = {}
    for _, field in ipairs(raw_fields) do
        if #field.value == 1 then
            msg[field.name] = field.value[1]
        else
            msg[field.name] = field.value
        end
    end

    -- scrub agent
    for field, prefix in pairs(cfg.user_agent_transforms) do
        if msg[field] then
            msg[prefix.."browser"],
            msg[prefix.."version"],
            msg[prefix.."os"] = clf.normalize_user_agent(msg[field])
            if not ((cfg.user_agent_conditional and not msg[prefix.."browser"]) or cfg.user_agent_keep) then
                msg[field] = nil
            end
        end
    end

    orig = msg
    msg = {Fields={}}
    for key, value in pairs(orig) do
        if key:sub(1,7) == "Fields." then
            msg[key] = nil
            key = key:lower():gsub("[.-]","_")
            msg[key] = value
            if type(value) == 'table' then
                msg.Fields[key:sub(8, #key)] = concat(value, ",")
            else
                msg.Fields[key:sub(8, #key)] = tostring(value)
            end
        else
            key = key:gsub("[.-]","_")
            msg[key] = value
        end
    end
</pre>

relud:
This isn't a generic use case, closing wontfix

PR #81, opened 2017-02-17
Optionally format s3_path_dimensions using os.date

relud:


parquet/sandboxes/heka/output/s3_parquet.lua
```diff
@@ -112,6 +114,8 @@ require "parquet"
 local load_schema = require "lpeg.parquet".load_parquet_schema
 require "string"
 require "table"
+local date          = require "os".date
+local floor         = require "math".floor
```
trink: You don't need floor in this case

parquet/sandboxes/heka/output/s3_parquet.lua
```diff
@@ -242,6 +246,9 @@ local function get_s3_path(json)
         else
             v = read_json(json, d.source) or default_nil
         end
+        if d.dateformat and type(v) == "number" then
```
trink: So this allows any field to contain dateformat but requires it is in nanoseconds.  We could either restrict it to Timestamp only and dump the number test or we allow a scaling factor (say there is a time_t in some other field)

relud: scaling factor for sure

parquet/sandboxes/heka/output/s3_parquet.lua
relud: I'm using it on line 250, and without it os.date() will throw the error: `bad argument #2 to 'date' (number has no integer representation)`

parquet/sandboxes/heka/output/s3_parquet.lua
relud: I'll add a scaling factor that defaults to nanoseconds

parquet/sandboxes/heka/output/s3_parquet.lua
trink: line 120 can be removed along with the floor call on 252

PR #85, opened 2017-02-21
enable the kafka plugin in build/run.sh

relud:
requires mozilla-services/hindsight#86


relud:
nevermind, doing this differently

Issue #87, opened 2017-02-24
Elasticsearch Bulk API Output error handeling issues


champtar:
Hi @trink 

https://github.com/mozilla-services/lua_sandbox_extensions/blob/0fe1704dd079ae96f7f1be7402112eebf613876c/elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua

Multiple problems:

1) Using elasticsearch_bulk_api.lua with ```discard_on_error = false```,
if ```send_batch()``` return -1 or -2, we don't discard, but we don't set ```retry = true```, so we keep consuming message (I ended up with a 14G batch file :) )
We should retry or discard

2) if we hit a network error, ie ```send_batch()``` returns -3, we set ```retry = true```.
Now in a future ```process_message()``` call, if ```send_batch()``` returns 0, we just skipped/dropped a message? (not sure about this one)

3) The retry errors are logged at debug level
https://github.com/mozilla-services/hindsight/blob/7c66c4bd1043366cba1a38f056bc6630b2bfb736/src/hs_output_plugins.c#L429
Is there a way to log from a sandbox at a different level, without killing the sandbox with error (having debug/info/notice/warning would be great)?
Other solutions are:
- always log err at warning level if ret < 0
- add an optional third return param, ret|err|level
ie I don't want the "input_reader analysis0 opened file" messages, but I do want "output.elasticsearch retry message X err: connection refused"

4) discard never log the fact that it discarded a batch, and we can't increment the fail counter by the batch size

5) on restart we don't reread the batch_count

I'll try to send a PR for the simple issues, but I think we might need a separate process_batch api, because we can't really return meaningful return code when we do 2 totally separate actions

Your thoughts?

P.S: Thanks a lot for hindsight, it really rocks, fast, low memory usage, stable

trink:
1) IIRC it was by design; the thought was we would cache and not block the pipeline while elastic search was unavailable.  If it never comes back there aren't a lot of options you could terminate the plugin and optionally stop Hindsight.  An infinite retry would eventually backpressure the entire system and the reason for not discarding is because that is how it was configured.

2) No the retry applies to the message already added to the batch so no data will be lost when it resumes.

3)  Only a failure return (-1) will log a warning

4) Correct there isn't any mechanism to feed that information back at the moment.

5) If you have set 'preserve_data = true' set then batch_count is restored along with the retry state

trink:
[...]

PR #88, opened 2017-02-24
Rework elasticsearch_bulk_api.lua

champtar:
Fix some issues from #87


trink:
See comments in #87

champtar:
Quick question, is batch:close() really needed?

trink:
In shutdown, not really as the GC will close it.

PR #91, opened 2017-02-27
elasticsearch_bulk_api: rework error handling

champtar:
You can now choose to discard_on_error or abort_on_error,
after retrying max_retry.
If both discard_on_error and abort_on_error are false, retry forever

reopenning against dev branch

### DIFF
--- a/elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
+++ b/elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
@@ -17,8 +17,12 @@ port                = 9200
 timeout             = 10
 flush_count         = 50000
 flush_on_shutdown   = false
-preserve_data       = not flush_on_shutdown --in most cases this should be the inverse of flush_on_shutdown
-discard_on_error    = false
+-- preserve_data       = false -- we don't rely on preserve data anymore
+max_retry           = 0
+discard_on_error    = false -- discard the batch after max_retry+1 failed attemps to send the batch
+abort_on_error      = false -- stop this plugin after max_retry+1 failed attemps to send the batch
+-- when setting abort_on_error=true, consider also settings shutdown_on_terminate or remove_checkpoints_on_terminate
+
 
 -- See the elasticsearch module directory for the various encoders and configuration documentation.
 encoder_module  = "encoders.elasticsearch.payload"
@@ -41,6 +45,8 @@ local address   = read_config("address") or "127.0.0.1"
 local port      = read_config("port") or 9200
 local timeout   = read_config("timeout") or 10
 local discard   = read_config("discard_on_error")
+local abort     = read_config("abort_on_error")
+local maxretry  = read_config("max_retry") or 0
 
 local encoder_module = read_config("encoder_module") or "encoders.elasticsearch.payload"
 local encode = require(encoder_module).encode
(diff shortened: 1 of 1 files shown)


elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
```diff
@@ -144,6 +148,37 @@ end
...
+local function send_batch2()
+    try_count = try_count + 1
+    local ret, err = send_batch()
+    if ret < 0 then client = nil end -- always use a new connection after an error
+
+    if ret == 0 then
+        retry = false
+    elseif discard and try_count > maxretry then
+        retry = false
+        err = "Discarding "..batch_count.." messages after "..try_count.." attemps, ret="..ret.." err="..err
```
trink: Use string.format for all of these error messages

elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
```diff
@@ -144,6 +148,37 @@ end
...
+        ret = -1
+    elseif abort and try_count > maxretry then
+        retry = true -- not saved by preserve_data for now
```
trink: This does nothing since terminating the plugin will not save the state

elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
```diff
@@ -144,6 +148,37 @@ end
...
+        error("Abort sending "..batch_count.." messages after "..try_count.." attemps, ret="..ret.." err="..err)
```
trink: Now that it terminates it needs proper cleanup so the next restart won't send a double size batch (clear out the batch or re-establish the batch_count on restart after termination)

champtar: I messed up in my testing and though it was saving state more often (after last succesfull process_message())
What do you think about adding a new return code, like ```LSB_HEKA_PM_ABORT``` to do clean abort?

champtar: 2 small questions
-how can I stop Hindsight and not just the plugin? 
-at what point are the global variable (like retry) restored?

trink: See `shutdown_on_terminate`
https://github.com/mozilla-services/hindsight/blob/master/docs/configuration.md#default-sandbox-configuration-variable

They are saved when a plugin is cleanly stopped and are restored when the plugin is initialized before starting.

champtar:
Hi @trink,
I've updated my patch.
I'm only deleting the batch when discard_on_error == true because either the ES cluster was in a bad state, and we can ingest this batch later, or the data is malformed and keeping it give a chance to investigate.
I'm not using preserve_data anymore so i don't have a special case for abort

elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
```diff
@@ -142,8 +148,46 @@ local function send_batch()
...
-batch_count = 0
-retry       = false
+local batch_count = 0
+local retry       = false
+local try_count   = 0
+
+for _ in io.lines(batch_file) do
+    batch_count = batch_count + 1
+end
+batch_count = batch_count / 2
```
bbinet: Why do we need this division? Elasticsearch batch data takes 2 lines per message?

bbinet: It seems to be the case indeed: https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/elasticsearch/io_modules/encoders/elasticsearch/payload.lua

PR #101, opened 2017-03-15
Bug 1346243 Add normalizedChannel to telemetry.error messages

trink:


mreid-moz:
For consistency, it seems like we should either always use the channel (and hence `normalizedChannel`) from the URL or from the document. Since the document parse might fail, that leaves the URL as the most reliable source.

Is there an easy way to check if we ever observe a difference between the URL component and the channel in the document?

whd:
While the document parse might fail, https://bugzilla.mozilla.org/show_bug.cgi?id=1339024 demonstrates that the URL can also sometimes contain garbage. In some cases we might have a valid channel in the document but not in the URL (and v.v.).

mreid-moz:
Ok, then my preference is to use the URL Component to set an initial value for `normalizedChannel`, but to continue overriding it (and `appUpdateChannel`) with the in-document value wherever available.

PR #106, opened 2017-03-27
Add the pingSenderVersion field to the decoder

Dexterp37:
This will enable storing the pingsender version that
comes from the headers of the pings sent with it.


Dexterp37:
Closing this PR, as this should be against /dev not /master.

PR #107, opened 2017-03-27
Add the pingSenderVersion field to the decoder

Dexterp37:
This will enable storing the pingsender version that
comes from the headers of the pings sent with it.

### DIFF
--- a/moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
+++ b/moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
@@ -478,6 +478,7 @@ function transform_message(hsr)
         msg.Fields.geoCity         = hsr:read_message("Fields[geoCity]")
         msg.Fields.submissionDate  = os.date("%Y%m%d", hsr:read_message("Timestamp") / 1e9)
         msg.Fields.sourceName      = "telemetry"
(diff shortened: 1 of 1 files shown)


moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
```diff
@@ -478,6 +478,7 @@ function transform_message(hsr)
         msg.Fields.geoCity         = hsr:read_message("Fields[geoCity]")
         msg.Fields.submissionDate  = os.date("%Y%m%d", hsr:read_message("Timestamp") / 1e9)
         msg.Fields.sourceName      = "telemetry"
+        msg.Fields.pingSenderVersion = hsr:read_message("Fields[X-PingSender-Version]")
```
trink: I am not big on changing the name.  I would prefer if the key stayed 'X-PingSender-Version'
@mreid-moz  may feel differently

mreid-moz: The other headers retain their original names, so +1 on keeping it `X-PingSender-Version`.

Dexterp37: Good point! I changed the code, I hope the LUA syntax is still correct.

mreid-moz:
@trink does this look OK now?

PR #120, opened 2017-04-25
Less restrictive values for HTTP methods

fvanniere:
HTTP method in logfiles can by any uppercase word with more than 3 letters.


trink:
The grammar above is also too restrictive; beyond the standard methods a token is accepted see:
https://tools.ietf.org/html/rfc7231#page-21
https://tools.ietf.org/html/rfc7230#section-3.2.6

Please update and re-open the PR against the dev branch as per https://mail.mozilla.org/pipermail/hindsight/2017-January/000043.html (I will add that information into the README) or file an issue to get the grammar updated.

Thanks,
Trink

PR #126, opened 2017-04-28
Bugzilla 1337927 - Add support for loading versioned schemas

trink:
r? @mreid-moz

### DIFF
--- a/moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
+++ b/moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
@@ -8,8 +8,13 @@
 ## Decoder Configuration Table
 ```lua
 decoders_moz_telemetry_ping = {
-    -- String used to specify the schema location on disk.
-    schema_path = "/mnt/work/schemas",
+    -- String used to specify the schema location on disk. The path should
+    -- contain one directory for each docType and the files in the directory
+    -- must be named <docType>.<version>.schema.json. If the schema file is not
+    -- found for a docType/version combination, the default schema is used to
+    -- verify the document is a valid json object.
+    -- e.g., main/main.4.schema.json
+    schema_path = "/mnt/work/mozilla-pipeline-schemas/schemas/telemetry",
 
     -- String used to specify the message field containing the user submitted telemetry ping.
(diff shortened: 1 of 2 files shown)


moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
```diff
@@ -323,26 +337,35 @@ local function remove_objects(msg, doc, section, objects)
 end
 
 
+local function validate_schema(hsr, msg, doc, version)
```
trink: The vacuous schema is no different than any other (whatever file is provided by the schema service is used).  The only difference is if no schema match is found it is used by default.  Would you prefer the schema be baked into this module?

trink: There are several way this can be accomplished but until the schema deployment mechanism is defined it would be premature.   Currently when new schemas are deployed the plugins are restarted meaning they are automatically re-loaded.

mreid-moz: re: vacuous schema - the behaviour is fine as-is, I'd just like to see a descriptive comment for future readers about how it is used.

Sounds good re: schema reloading.

trink: Built the default_schema in and documented its behavior; 'vacuous' can be removed from the schema repo

moz_telemetry/io_modules/decoders/moz_telemetry/ping.lua
```diff
@@ -323,26 +349,39 @@ local function remove_objects(msg, doc, section, objects)
...
+    local dt = schemas[msg.Fields.docType or ""]
+    if dt then
+        version = tonumber(version)
+        if not version then version = 1 end
+        schema = dt[version] or default_schema
+    end
+
+    ok, err = doc:validate(schema)
+    if not ok then
+        inject_error(hsr, "json", string.format("%s schema validation error: %s", msg.Fields.docType, err), msg.Fields)
```
mreid-moz: Please add the `version` field to the error message string as well.

PR #143, opened 2017-06-12
Sprint May 29

trink:


moz_pioneer/io_modules/decoders/moz_pioneer/jose.lua
```diff
@@ -0,0 +1,342 @@
...
+        return false
+    end
+
+    ok, err = doc:validate(schema)
+    if not ok then
+        inject_error(hsr, "json", string.format("%s schema version %d validation error: %s", msg.Fields.docType, msg.Fields.sourceVersion, err), msg.Fields)
+        return false
+    end
+
+    submissionField.value = doc
```
trink: As an optimization I would recommend removing the JSON encoding entirely and expanding the structure into the native protobuf message.

PR #145, opened 2017-06-13
Fix the data loss in the ElasticSearch output when ES is down

trink:


beorn-:
Sure this PR fixes messages drops. the scenario tested was:
- inject 1 message
- check if ther'es 1 message
- shutdown ES cluster
- inject 2 messages
- start ES cluster
- check if there's now 3 messages
- inject 1 message
- check if there's now 4 messages

Went fine here :+1:

champtar:
Hi @trink 
Can't find this commit into master or dev, was it dropped?

trink:
Sigh, It was dropped in a rebase of the June 12 sprint branch. I added it back into the current sprint e3b63038232970238cf4123badf4dfc151517767

champtar:
Also in what case were you loosing messages, because i just tested with a ~1month old build, and I'm not able to loose any messages (only to have duplicates if the timeout is to small)

My tests, with a 3 nodes ES cluster (A/B/C), hindsight sending to A
stopping A, sending 100 logs, restarting A, i gain at least 100 logs
stopping B&C, sending 100 logs, restarting B&C, i gain at least 100 logs

Regards

trink:
Beorn's test case demonstrated the failure https://github.com/mozilla-services/lua_sandbox_extensions/pull/145#issuecomment-308157972

There was also a Hindsight retry issue https://github.com/mozilla-services/hindsight/commit/75c28345433457c15ec59b522f5148841c379e69:
- stop ES
- send message 1 (HS ES output is in a retry state)
- stop HS
- start HS
- send message 2 (ES would only receive this message)

champtar:
Right now I'm using:
- https://github.com/mozilla-services/hindsight/commit/ade19dda17bce2330cb29a657ca772507aebbf1d
- https://github.com/mozilla-services/lua_sandbox/commit/125c18431eb520366ce75854fce1e88cd0e2226b
- https://github.com/mozilla-services/lua_sandbox_extensions/commit/9abb0f9c8b6c38ed982cd2526f9f0cbde11c4542

And I'm unable to reproduce this log loss ...

I will update everything to latest master and try again

beorn-:
@champtar there's one tiny precision missing here, whenever i was stopping completely ES it was behaving OK.

But whenvever the ES cluster was behind a HTTP reverse proxy (i.e. haproxy), haproxy was still answering with 5xx http error codes instead of leading to a l4 connection problem. which made the difference here.

PR #146, opened 2017-06-15
lpeg: Add logfmt lpeg grammar and hindsight lua decoder

beorn-:
This PR adds a basic support for logfmt format and a decoder that allows remaping and casting fields

### DIFF
--- a/heka/CMakeLists.txt
+++ b/heka/CMakeLists.txt
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
-project(heka VERSION 1.1.10 LANGUAGES C)
+project(heka VERSION 1.1.11 LANGUAGES C)
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Utility modules for Heka sandboxes")
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "${PACKAGE_PREFIX}-cjson (>= 2.1)")
 string(REGEX REPLACE "[()]" "" CPACK_RPM_PACKAGE_REQUIRES ${CPACK_DEBIAN_PACKAGE_DEPENDS})
--- a/lpeg/io_modules/decoders/logfmt.lua
+++ b/lpeg/io_modules/decoders/logfmt.lua
@@ -1,0 +1,135 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Logfmt Message Decoder Module
+
+## Decoder Configuration Table for input cfg file
+decoder_module = "decoders.logfmt"
+
+-- Allows you to remap parsed logfmt tags into another field
+logfmt_conf = {
+    -- override root fields using Fields fields
+    remap_hash_root = {
+        time   = "Timestamp",
+        host   = "Hostname",
+        pid    = "Pid",
+    },
+    -- rename Fields fields
+    remap_hash_fields = {
+        app        = "Program",
+        lvl        = "Severity",
+        level      = "Severity",
+        env        = "Environment",
+        msg        = "Message",
+        err        = "Error",
+        from       = "From",
+    },
+    -- key -- cast function to apply on root fields after remaps
+    cast_hash_root = {
+        Pid         = "tonumber",
+    },
+    -- key -- cast function to apply on Fields fields after remaps
+    cast_hash_fields = {
+        ok         = "tonumber",
+        ko         = "tonumber",
+        queued     = "tonumber",
+        discarded  = "tonumber",
+        alert      = "tonumber",
+        len        = "tonumber"
+    }
+}
+
+## Functions
+
+### decode
+
+Decode and inject the resulting message
+
+*Arguments*
+- data (string) - logfmt message 
+
+*Return*
+- nil - throws an error on an invalid data type inject_message failure etc.
+
+--]]
+
+-- Read configuration
+local cfg = read_config("logfmt_conf") or {}
+cfg.remap_hash_root = cfg.remap_hash_root or {}
+cfg.remap_hash_fields = cfg.remap_hash_fields or {}
+cfg.cast_hash_root = cfg.cast_hash_root or {}
+cfg.cast_hash_fields = cfg.cast_hash_fields or {}
+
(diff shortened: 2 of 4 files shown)


lpeg/io_modules/decoders/logfmt.lua
```diff
@@ -0,0 +1,135 @@
...
+--[[
+# Logfmt Message Decoder Module
+
+## Decoder Configuration Table for input cfg file
+decoder_module = "decoders.logfmt"
+
+-- Allows you to remap parsed logfmt tags into another field
+logfmt_conf = {
+    -- override root fields using Fields fields
+    remap_hash_root = {
```
trink: Some data best practices:
- It is generally a bad idea to alter the schema of an incoming message, if the schema needs to be changed it should be changed at the source.  Altering the schema breaks user expectations e.g.,  if I logfmt `level` as one of my key value pairs I wouldn't expect to look for it in msg.Severity. 
- Heka message headers are best used for metadata; in general avoid promoting data from a known schema format
- Not preserving type information is a failing of logfmt itself but mapping it in the cfg will just be a world of maintenance pain. Since logfmt is just a bag of key/value pairs and any app can log anything they want and change it at any time I would recommend something a little more dynamic.  Use the grammar to infer the type (at least number/string but you could also go deeper and include boolean/integer (although not perfect as one value may be foo = 1 and the the next message may be foo = 1.1 giving you two different schemas for the same type of message).  So instead of a manual mapping; settings like infer_type = true, numeric_type = "auto|double|integer" would be more flexible

lpeg/io_modules/decoders/logfmt.lua
```diff
@@ -0,0 +1,135 @@
...
+
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+if cfg.remap_hash then
+    print("decoders.logfmt: remap hash found")
+    for tag, target in pairs(cfg.remap_hash) do print("decoders.logfmt: ", tag," => ", target) end
+end
+
+function decode(data)
+    local msg = { }
```
trink: Move this outside of decode to avoid some GC

lpeg/io_modules/decoders/logfmt.lua
```diff
@@ -0,0 +1,135 @@
...
+    local mapped = false
+    local fields = logfmt:match(data)
+    if not fields then return "parse failed" end
+
+    msg.Type = "logfmt"
+    msg.Payload = data
```
trink: Preserving the original message is usually a cfg option `payload_keep`

lpeg/io_modules/decoders/logfmt.lua
```diff
@@ -0,0 +1,135 @@
...
+
+    msg.EnvVersion = "1.0"
+    msg.Fields = {}
+
+    for k,v in pairs(fields) do
+        mapped = false
+        -- rename fields at Root level
+        if cfg.remap_hash_root[k] then
```
trink: Save the result to a local and avoid multiple lookups.

lpeg/modules/lpeg/logfmt.lua
```diff
@@ -0,0 +1,19 @@
+local l = require "lpeg"
```
trink: Add documentation

lpeg/modules/lpeg/logfmt.lua
```diff
@@ -0,0 +1,19 @@
...
+l.locale(l)
+
```
trink: Wrap in its own environment and return the table

lpeg/modules/lpeg/logfmt.lua
```diff
@@ -0,0 +1,19 @@
...
+local escdoublequote = l.P('\\"')
+local underscore = l.P("_")
+local anyletter = l.R("az")
+
+local key = l.C((anyletter + underscore)^1) * whitespace^0
+local val = l.C((1-(whitespace + doublequote))^1) + (doublequote * l.C((escdoublequote + (1 - doublequote))^1) * doublequote)
+
+local separator = whitespace^1
+local pair = l.Cg(key * "=" * (val + empty) ) * separator^-1
+grammar = l.Cf(emptytable * pair^0, rawset)
```
trink: Add unit tests

lpeg/io_modules/decoders/logfmt.lua
```diff
@@ -0,0 +1,135 @@
...
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Logfmt Message Decoder Module
+
+## Decoder Configuration Table for input cfg file
+decoder_module = "decoders.logfmt"
+
+-- Allows you to remap parsed logfmt tags into another field
+logfmt_conf = {
```
beorn-: done

heka/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
-project(heka VERSION 1.1.10 LANGUAGES C)
+project(heka VERSION 1.1.11 LANGUAGES C)
```
beorn-: done

lpeg/io_modules/decoders/logfmt.lua
beorn-: Even if i clearly am agreeing with you here, real life shows that you sometimes can't get people/code to conform to these ideals.

To me it has proved handy and even mandatory in sadly too many setups. I'm even adding this feature in a few lua plugins.

But if you prefer not to encourage these practices by dropping this feature it's fine by me.

wdyt ?

lpeg/io_modules/decoders/logfmt.lua
beorn-: done

lpeg/io_modules/decoders/logfmt.lua
beorn-: added

lpeg/io_modules/decoders/logfmt.lua
trink: I would prefer not to encourage fixing-up/altering schemas in real-time (as opposed to fixing it at the source, I also understand that fixing the source is not always possible).  However, we have been actively removing mandatory restrictions/hard limits, in Hindsight, and simply providing recommendations to allow this flexibility when needed... so lets keep it but label it with a warning/caveats.  Also, you say you are adding similar functionality in multiple places?  Can it be factored out and put in 'transformation' module?  Or at least share a common naming conventions/structure?

trink:
Replaced by #212 and #211

PR #154, opened 2017-07-31
Encode NULLs in maps as parser.null

washort:


parquet/io_modules/lpeg/parquet.lua
```diff
@@ -216,6 +216,8 @@ function load_parquet_schema(spec, hive_compatible, metadata_group)
     return root, f
 end
 
+null = parquet.null
```
trink: why define another null variable alias in the grammar?

parquet/parquet.cpp
```diff
@@ -1497,6 +1497,10 @@ static void dissect_field(pq_writer *pw, lua_State *lua, pq_node *n, int16_t r,
       }
     }
     break;
+  case LUA_TLIGHTUSERDATA:
+    if (lua_touserdata(lua, -1) != NULL) {
+      break;
```
trink: This will create an invalid parquet file if a non-null lightweight user data objects is passed

parquet/parquet.cpp
```diff
@@ -1497,6 +1497,10 @@ static void dissect_field(pq_writer *pw, lua_State *lua, pq_node *n, int16_t r,
...
+    }
```
trink: comment the intended fall thru (GCC 7 is now making this a warning and it will fail the build)

trink:
The output plugin has to be modified to enable NULL support `cjson.decode_null(true)`
https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/parquet/sandboxes/heka/output/s3_parquet.lua#L135

trink:
Also, you will need to bump the version in the CMakeLists.txt and the test.lua

parquet/io_modules/lpeg/parquet.lua
washort: Misunderstood how the files/modules were organized.

parquet/sandboxes/heka/output/s3_parquet.lua
```diff
@@ -130,9 +130,13 @@ local hindsight_admin       = read_config("hindsight_admin")
 local hostname              = read_config("Hostname")
 local metadata_group        = read_config("metadata_group")
 local json_objects          = read_config("json_objects")
+local json_decode_null      = read_config("json_decode_null")
 local json_objects_len      = 0
 if type(json_objects) == "table" then
     require "cjson"
+    if json_decode null then
```
washort: See also https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/71

washort:
I believe this is complete now; I didn't understand how repetition levels worked, but now parquet-cpp and parquet-mr commandline tools indicate this is outputting nulls properly.

parquet/parquet.cpp
```diff
@@ -1356,7 +1356,7 @@ static void add_null(pq_column *c, int16_t r, int16_t d)
     ss << "column '" << c->n->name << "' is required";
     throw pq::ParquetException(ss.str());
   }
-  update_levels(c, r, d);
+  update_levels(c, r, d - 1 ? d > 0 : 0);
```
trink: This breaks the dremel depth encoding on nested structures.

washort: My understanding (reflected by tools like `parquet_reader` and `parquet-dump`) is that the way to encode nulls is to use the definition level for the next level up. So using the test schema here ( https://github.com/washort/lua_sandbox_extensions/blob/3a1d6b63b5605905f7b09b02ae923de8b5ac26cb/parquet/tests/test.lua#L246-L270 ) an input like `{my_map = {foo = 1, bar = parquet.null, blee = 2}}` results in these parquet columns:

```
BINARY my_map.key_value.key 
----------------------------------------------------------------------------------------------------------------------------------------
*** row group 1 of 1, values 1 to 3 *** 
value 1: R:0 D:1 V:blee
value 2: R:1 D:1 V:bar
value 3: R:1 D:1 V:foo

INT32 my_map.key_value.value 
----------------------------------------------------------------------------------------------------------------------------------------
*** row group 1 of 1, values 1 to 3 *** 
value 1: R:0 D:2 V:2
value 2: R:1 D:1 V:<null>
value 3: R:1 D:2 V:1
```

That's from `parquet-dump`, here's the `parquet_reader_` output: https://gist.github.com/washort/41a31c71adf5899d1e2d5976218d4a30

trink: See the example in https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/36632.pdf.  The example output is broken in this PR.

parquet/parquet.cpp
```diff
@@ -1497,6 +1497,12 @@ static void dissect_field(pq_writer *pw, lua_State *lua, pq_node *n, int16_t r,
       }
     }
     break;
+  case LUA_TLIGHTUSERDATA:
+    if (lua_touserdata(lua, -1) != NULL) {
+      add_value(lua, n, c, r, n->dl);
+      break;
+    }
+    /* fall through */
```
trink: Use the special syntax `/* FALLTHRU */` see https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

parquet/parquet.cpp
washort: Tricky stuff.

I saw the breakage you describe in `example.parquet`. While investigating this I discovered that lists were encoding nulls improperly as well. I don't know if there's a more general solution needed but the use of `maybe_dissect_field`in `dissect_list` and `dissect_map` satisfies the cases I tested.

parquet/parquet.cpp
```diff
@@ -1523,7 +1536,7 @@ static void dissect_map(pq_writer *pw, lua_State *lua, pq_node *n, int16_t r, in
   bool found = false;
   lua_pushnil(lua);
   while (lua_next(lua, -2) != 0) {
-    dissect_field(pw, lua, vn, cr, vn->dl);
```
trink: the following will resolve the issue, dissect just needs the proper parent node depth since we don't have a guaranteed leaf node anymore. : 
`dissect_field(pw, lua, vn, cr, n->dl);`

parquet/parquet.cpp
```diff
@@ -1544,7 +1557,7 @@ static void dissect_list(pq_writer *pw, lua_State *lua, pq_node *n, int16_t r, i
...
```
trink: same here:
`dissect_field(pw, lua, vn, cr, n->dl);`

parquet/parquet.cpp
washort: Fixed.

PR #155, opened 2017-08-03
test fixes

washort:


trink:
The GCC7 build issues will be handled when I get back as it touches a lot more than this.  Also you can avoid the cmake changes by simply using the specified comment syntax as per https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html 
    `/* FALLTHRU */`

Also all PRs should be made against the dev branch.

PR #159, opened 2017-08-08
Use patterns to make jose test pass

ishigoemon:
With debian jessie-backports libcjose-dev:0.4.1-3~bpo8+1, jose tests fail due to different line numbers referenced in the error messages. This PR changes those tests to be pattern matches with the line number matched to %d+. Other tests were left as string comparisons to avoid escaping ?'s and ()'s.

### DIFF
--- a/jose/tests/test.lua
+++ b/jose/tests/test.lua
@@ -32,11 +32,11 @@ local bad = jose.header({alg = "foo", enc = "A256GCM"})
 local construction_errors = {
     {fn = jose.jwk_import, args = {}        , err ="bad argument #0 to '?' (incorrect number of arguments)"},
     {fn = jose.jwk_import, args = {true}    , err ="bad argument #1 to '?' (string expected, got boolean)"},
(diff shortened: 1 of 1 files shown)


jose/tests/test.lua
```diff
@@ -61,12 +61,16 @@ local construction_errors = {
...
     {fn = jose.jws_sign, args = {jwk, payload, bad},
-        err = "file: jws.c line: 112 function: _cjose_jws_validate_hdr message: invalid argument"},
+        perr = "file: jws.c line: %d+ function: _cjose_jws_validate_hdr message: invalid argument"},
 }
 
 for i, v in ipairs(construction_errors) do
     local ok, err = pcall(v.fn, unpack(v.args))
-    if err ~= v.err then error(string.format("error test %d failed %s\n", i, err)) end
+    if v.perr then
+        if not string.find(err, v.err) then error(string.format("error test %d failed %s\n", i, err)) end
```
trink: string.match is sufficient and you would want to use v.perr for the pattern

PR #160, opened 2017-08-21
Bugzilla 1368197 Implement generic ingestion decoder

whd:
This is mostly code duplication / simplification of the telemetry decoder.

Of note: this implementation doesn't support additional (optional) path components. We sort of spec'd them but don't have a good way of generically specifying them. In theory we could add path components to the schemas repo somewhere and have `uri_config` be generated from it but for now it's simply an error. We also propogate `Fields[args]` from the the edge message without any processing.

I'm happy to rename things; generic.generic is a bit silly.

The build is failing on Ubuntu and Fedora for an unrelated GCC7 issue.


moz_generic/tests/hindsight/run/output/d2p.cfg
```diff
@@ -0,0 +1,29 @@
+filename        = "s3_parquet.lua"
+message_matcher = "Type == 'moz_generic' && Fields[namespace] == 'generic' && Fields[docType] == 'test' && Fields[docVersion] == '1'"
+ticker_interval = 60
+preserve_data   = true
+
+parquet_schema = [=[
```
trink: Technically all the input specific parquet schemas would have already been tested here https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/tests/hindsight/run/output/test_parquet_schemas.lua what is the point of writing out a generic parquet file?  Wouldn't we want a generic_s3_parquet output?

moz_generic/tests/test.lua
```diff
@@ -0,0 +1,6 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
trink: See related comment above (remove test dependency on no module packages)

moz_generic/test_sandbox.c
```diff
@@ -0,0 +1,51 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
```
whd: Is this actionable in this PR?

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ kafka
 lfs
 lpeg
 lsb
+moz_generic
```
whd: I've switched the base to dev.

See [this google doc](https://docs.google.com/document/d/1PqiF1rF2fCk_kQuGSwGwildDf4Crg9MJTY44E6N5DSk/edit?ts=5910c4cf) in addition to [these](https://bugzilla.mozilla.org/show_bug.cgi?id=1368197) [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1368196) for some specs.

Generic ingestion is essentially telemetry without all the edge case logic. The theory is that we could support other input and output schemas eventually, it's just that we don't have any use cases that aren't json at the moment.

While we could implement this as an extension of the telemetry decoder, I think it would result in more complicated code with a high chance of causing unexpected regressions. I would prefer to maintain a simpler, separate decoder that cannot affect telemetry processing to changing the current decoder. @mreid-moz may have an opinion on this as well.

If we do want to maintain a single decoder, I would be most comfortable with some kind of top-level flag applied to decoders performing generic ingestion that makes the branching logic very apparent while still allowing us to avoid code duplication.

moz_generic/io_modules/decoders/moz_generic/generic.lua
```diff
@@ -0,0 +1,438 @@
...
+## Decoder Configuration Table
+```lua
+decoders_moz_generic = {
+    -- String used to specify the schema location on disk. The path should
+    -- contain directories for each namespace and docType and the files in
+    -- the namespace directories must be named <docType>.<version>.schema.json.
+    -- e.g., generic/testping/testping.4.schema.json
+    schema_path = "/mnt/work/mozilla-pipeline-schemas/schemas",
+
+    -- String used to specify the message field containing the user submitted payload.
```
whd: Removed.

moz_generic/io_modules/decoders/moz_generic/generic.lua
```diff
@@ -0,0 +1,438 @@
...
+    content_field = "Fields[content]", -- optional, default shown
+
+    -- String used to specify the message field containing the URI of the submission.
```
whd: Removed.

moz_generic/tests/hindsight/run/output/d2p.cfg
whd: I've removed this test. It was useful for me when developing, but it indeed does not really test anything new.

As for a generic_s3_parquet output, that was the plan, but per https://bugzilla.mozilla.org/show_bug.cgi?id=1368197#c4 I think it's simpler to dynamically generate the hindsight configs and use the existing output than to write a multiplexed parquet output.

moz_generic/test_sandbox.c
trink: After you rebase I can add the change to this branch

CMakeLists.txt
trink: Was thinking of sub-decoder support like in the syslog decoder

trink:
This will be re-designed to handle all output of nginx_moz_ingest (including telemetry)

PR #169, opened 2017-09-20
Bug 1381954: Filter out non-whitelisted experiment types and blocklisted experimen…

sunahsuh:
…t ids in moz_telemetry_s3

I was able to get the integration tests running with a hacked-together Dockerfile (I'm considering how to generalize this and contribute it back but it'll require a little more work than I can put in at the moment.)

I have no idea where the production `moz_telemetry_s3.cfg` lives so I can add experiment_types and experiment_blocklist, but if you can point me in that direction I'll submit a PR for that as well.

### DIFF
--- a/moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
+++ b/moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
@@ -49,6 +49,10 @@ preserve_data       = not flush_on_shutdown -- should always be the inverse of f
 
 -- Specify an optional module to encode incoming messages via its encode function.
 -- encoder_module = "encoders.heka.framed_protobuf" -- default
+
+-- Specifies experiment types whitelist, and experiment id blocklist
+experiment_types = {["normandy-preference-"] = true} -- optional
+experiment_blocklist = {["pref-flip-screenshots-release-1369150"] = true} -- optional
 ```
 --]]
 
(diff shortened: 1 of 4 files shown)


moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
```diff
@@ -85,6 +85,10 @@ if experiment_dimensions then
     experiment_batch_dir  = read_config("experiment_batch_dir") or error("experiment_batch_dir must be specified")
     experiment_dimensions = mts3.validate_dimensions(experiment_dimensions)
 end
+local experiment_types = read_config("experiment_types")
+if not experiment_types then experiment_types = {} end
+local experiment_blocklist = read_config("experiment_blocklist")
+if not experiment_blocklist then experiment_types = {} end
```
trink: Sets the wrong value when the blocklist is nil/false;

Also this and the above statement allow the config option to be a non-table (which will fail below at run time instead of init time). Something like the following would be more robust:

```lua
local experiment_blocklist = read_config("experiment_blocklist") or {}
assert(type(experiment_blocklist) == "table", "experiment_blocklist must be a table")
```

moz_telemetry/sandboxes/heka/output/moz_telemetry_s3.lua
```diff
@@ -85,6 +85,10 @@ if experiment_dimensions then
     experiment_batch_dir  = read_config("experiment_batch_dir") or error("experiment_batch_dir must be specified")
     experiment_dimensions = mts3.validate_dimensions(experiment_dimensions)
 end
+local experiment_types = read_config("experiment_types")
```
trink: These should be added to the documentation

sunahsuh:
Thanks trink, RFAL

And, do you know, where I can make changes to the prod config file?

trink:
https://github.com/mozilla-services/puppet-config/blob/master/pipeline/modules/pipeline/templates/hindsight/output/telemetry_s3.cfg.erb

trink:
Looks good please bump the version number in moz_telemetry/CMakeLists.txt to 1.2.15

sunahsuh:
Version bumped, should be good to go!

PR #177, opened 2017-11-08
add 'normalizedOSVersion' to validated pings (bug #1401969)

washort:


moz_telemetry/modules/moz_telemetry/normalize.lua
```diff
@@ -192,6 +192,13 @@ function os(name)
     return normalize_os_grammar:match(name)
 end
 
+local normalize_os_version_grammar =
+l.C(l.R"09"^1 * (l.P"." * l.R"09"^1)^-1 * (l.P"." * l.R"09"^1)^-1)
```
trink: `l.C(l.R"09"^1 * (l.P"." * l.R"09"^1)^-2)` should suffice

moz_telemetry/modules/moz_telemetry/normalize.lua
```diff
@@ -192,6 +192,13 @@ function os(name)
...
+
+function os_version(v)
+    if type(v) ~= "string" then v = "" end
```
trink: why not return nil

trink:
Also, please bump the package version to 1.2.16

mreid-moz:
Can you rebase this on the latest `dev` to get rid of the noise in the commit history?

moz_telemetry/modules/moz_telemetry/normalize.lua
mreid-moz: Also note per [the schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/telemetry/environment.1.schema.json#L611-L613) that this could come in as an integer, in that case we should stringify it directly.

mreid-moz: @washort please update the code to handle the case where os version is a number.

washort:
rebased, review comments addressed

moz_telemetry/modules/moz_telemetry/normalize.lua
```diff
@@ -192,6 +192,15 @@ function os(name)
...
 end
 
+local normalize_os_version_grammar =
+    l.C(l.R"09"^1 * (l.P"." * l.R"09"^1)^-2)
+
+function os_version(v)
+    if type(v) == "string" then
+        return normalize_os_version_grammar:match(v)
+    else
+        return tostring(v)
```
mreid-moz: Please convert to a string value before running the 'normalize' grammar to avoid things like `-1` or `nil` leaking into the normalized field.

Something like:
```
vs = tostring(v)
if vs then
  return normalize_os_version_grammar:match(vs)
end
```

moz_telemetry/modules/moz_telemetry/normalize.lua
```diff
@@ -192,6 +192,15 @@ function os(name)
     return normalize_os_grammar:match(name)
 end
 
+local normalize_os_version_grammar =
+    l.C(l.R"09"^1 * (l.P"." * l.R"09"^1)^-2)
+
+function os_version(v)
+    vstr = tostring(v)
+    if vstr then
+        return normalize_os_version_grammar:match(v)
```
mreid-moz: We should be matching against `vstr` rather than `v` here.

mreid-moz:
This looks good to me now. @trink can you take another look?

moz_telemetry/modules/moz_telemetry/normalize.lua
```diff
@@ -192,6 +192,15 @@ function os(name)
     return normalize_os_grammar:match(name)
 end
 
+local normalize_os_version_grammar =
+    l.C(l.R"09"^1 * (l.P"." * l.R"09"^1)^-2)
+
+function os_version(v)
+    vstr = tostring(v)
```
trink: make this local

trink:
This change broke the integration tests and would break all production data?  How was it tested?

DecodeError type: 0 representation: <nil> value: ...r/lib/luasandbox/modules/moz_telemetry/normalize.lua:199: attempt to call global 'tostring' (a nil value)
@

PR #180, opened 2017-11-10
ip reputation output plugin

g-k:
refs: https://bugzilla.mozilla.org/show_bug.cgi?id=1403574

Should convert moz security fxa heavy hitters tsv output into a table for the [ip-reputation-lua-client](https://github.com/mozilla-services/ip-reputation-lua-client) client.send_violations to send.

@trink Is this the right place to put the file? How and where should I write unit tests?

I'm still running into trouble packaging the lua client, so any help would be appreciated too.

### DIFF
--- a/moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
+++ b/moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
@@ -1,0 +1,317 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+#  Mozilla Security Heavy Hitters Monitor to IP Reputation Output Plugin
+
+See: https://cacm.acm.org/magazines/2009/10/42481-finding-the-frequent-items-in-streams-of-data/abstract
+
+Computes heavy hitters by IP and send them to the IP Reputation Service (tigerblood) with HTTP with Hawk auth.
+
+## Sample Configuration
+```lua
+filename = "moz_security_hh_ip_reputation.lua"
+message_matcher = "Logger == 'input.nginx'"
+ticker_interval = 60
+preserve_data = true
+
+id_field = "Fields[remote_addr]"
+-- id_field_capture = ",? *([^,]+)$"a -- optional e.g. extract the last entry in a comma delimited list
+
+-- hh_items = 10000 -- optional, defaults to 10000 (maximum number of heavy hitter IDs to track), NB: should be <= tigerblood's MAX_ENTRIES setting
+
+-- config for publishing to IP Reputation
+config = {
+   base_url = "https://tigerblood.prod.mozaws.net", -- NB: no trailing slash
+   id = "fxa_heavy_hitters", -- hawk ID
+   _key = "8eb54d0469ef000f8819a248c3f4465dc506bb2ad339aaa09466644b2214ffc8", -- hawk secret
+}
+violation_type = "fxa:heavy-hitter-ip" -- IP reputation violation type to lookup reputation penalty for
+```
+
+--]]
+
+-- HH config
+local id_field = read_config("id_field") or error("id_field must be configured")
+local id_fieldc = read_config("id_field_capture")
+local hh_items = read_config("hh_items") or 10000
+
+-- IP Rep. output config
+local client_cfg = read_config("config") or error("No ip reputation client config specified")
+local violation_type = read_config("violation_type") or error("No ip  violation type specified")
+
+hh = {}
+hh_size = 0
+
+function process_message()
+    local id = read_message(id_field)
+    if not id then return -1, "no id_field" end
+    if id_fieldc then
+        id = string.match(id, id_fieldc)
+        if not id then return 0 end -- no error as the capture may intentionally reject entries
+    end
+
+    local cnt = hh[id]
+    if cnt then
+        hh[id] = cnt + 1
+    else
+        if hh_size >= hh_items then
+            for k, cnt in pairs(hh) do
+                cnt = cnt - 1
+                if 0 == cnt then
+                    hh[k] = nil
+                    hh_size = hh_size - 1
+                else
+                    hh[k] = cnt
+                end
+            end
+        else
+            hh_size = hh_size + 1
+            hh[id] = 1
+        end
+    end
+    return 0
+end
+
+
+-- Hawk signing
+
+local string = require("string")
+local table = require("table")
+local digest = require("openssl").digest
+local hmac = require("openssl").hmac
+local random = require("openssl").random
+local mime = require("mime")
+
+local function url_escape(s)
+   return s:gsub("+", "-"):gsub("/", "_"):gsub("=", "")
+end
+
+-- returns a url safe base 64 nonce
+local function nonce(size)
+   return url_escape(mime.b64(random(size)))
+end
+
+-- openssl supports more just sha256 for now
+local function supported_hash(algorithm)
+   return algorithm == "sha256"
+end
+
+-- hash the normalized hawk payload string
+local function hash_payload(payload, algorithm, content_type)
+   if not supported_hash(algorithm) then
+      error("Unsupported algorithm.")
+   end
+
+   local normalized = "hawk.1.payload"
+   if content_type then
+      normalized = string.format("%s\n%s", normalized, string.lower(content_type))
+   end
+   if payload then
+      normalized = string.format("%s\n%s", normalized, payload)
+   end
+   normalized = string.format("%s\n", normalized)
+
+   return mime.b64(digest.digest(algorithm, normalized, true)) -- digest w/ true to return binary
+end
+
+
+-- returns a normalized hawk header
+local function normalize_string(artifacts)
+   local normalized = {
+      "hawk.1.header",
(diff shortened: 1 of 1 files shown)


trink:
Sandboxes require integration tests (you have to generate the data and validate the output using the entire pipeline). The tests are in that repo  (the existing tests need to be revamped as the data and the business logic changed this week and all testing was done with live data and endpoints) https://github.com/mozilla-services/lua_sandbox_extensions/tree/master/moz_security/tests/hindsight

Not sure what you mean by packing the lua client (I assume you are doing something other than building/deploying the RPMs)..

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
```diff
@@ -0,0 +1,67 @@
...
+
+local cfg = read_config("config") or error("No ip reputation client config specified")
+local violation_type = read_config("violation_type") or error("No ip  violation type specified")
+
+client.configure(cfg)
+
+
+function process_message()
+   local pt = read_message("Fields[payload_type]")
+   if type(pt) ~= "tsv" then return -1, "invalid payload_type" end
```
trink: `type(pt)` will return "string". Use `pt ~= "tsv"`

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
```diff
@@ -0,0 +1,67 @@
...
+
+   local payload = read_message("Payload")
+
+   -- convert tsv to table for the client to convert to cjson
+   local violations = {}
+   local i = 0
+   for line in split(payload, "\n") do
```
trink: split is undefined

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
```diff
@@ -0,0 +1,67 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
trink: Also, I have no issues having the heavy hitters plugin output JSON instead of having to convert the tsv here.  Technically you don't even need the analysis plugin as you can do everything here (unless you want browse the output)

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
```diff
@@ -0,0 +1,67 @@
...
+--[[
+# Mozilla Security IP Reputation Output
+
+Sends tsv of IP and violation weights to the IP Reputation Service (tigerblood) with HTTP with Hawk auth.
+
+## Sample Configuration
+
+```lua
+filename = "moz_telemetry_crash_summary.lua"
+message_matcher = "Type == 'inject_payload'"
```
trink: Heavy hitters will not be the only thing injecting tsv payloads

g-k: What's an efficient way to only match the HH output?  Adding a filter on Logger like `&& Logger == 'analysis.hh_foo'`?

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
g-k: The tsv output is useful for debugging, so I'm fine leaving it as tsv.  If we run into perf issues we could change the analysis plugin or change the server to accept tsv.

g-k:
re: packaging the lua client

Does that belong in this repo? Also, ulfr was saying if it's like a 20 minute thing for you to do then you should go ahead and do that.

Otherwise, can you point me to an example of that? I think I added the luarocks paths to the hindsight io module path before, but that's probably a bad idea.

g-k:
re: integration tests

Do you usually write tests for output plugins that hit third party services? If so, how do you recommend recording or mocking out the API calls?

trink:
If anything we test against the real thing (see the Kafka tests. the manual setup means we rarely run them (only during module development)).  If you wanted to mock a out module just configure the module path to point to a mock version that has the test stubs you want./need.

moz_security/sandboxes/heka/output/moz_security_ip_reputation.lua
trink: Yeah you can match it by Logger (if you keep the analysis plugin).  Otherwise you will just consume the source data i.e. all the nginx log data `message_matcher = 'Type =~ \'nginx.access\'%'` and do everything in the output plugin.

trink:
I have no issues with the the ip-reputation-lua-client being a standalone repo.  Here are examples of how that is done https://github.com/mozilla-services/lua_sandbox_extensions/wiki/Third-Party-Sandbox-Extensions or if you would like to add it to the lua_sandbox_extensions repo and simply piggy back on the packaging infrastructure there.

g-k:
Made changes proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1403574#c6

* use an output plugin on the logs directly to skip the double encoding and message matching issues
* inline relevant parts of the ip reputation client and hawk header computation since it looks like the sandbox already includes the necessary dependencies

r? @trink

moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
```diff
@@ -0,0 +1,333 @@
...
+-- sets client config values; must be called before making requests
+-- example usage:
+--
+-- configure_client({
+--    base_url = "http://localhost:8080", -- tigerblood service url w/o slash
+--    id = "root", -- hawk ID
+--    key = "toor", -- hawk key
+-- })
+--
+function configure_client(new_client_config)
```
trink: configure_client and json request can be local

g-k:
Thanks! I made the following changes which should cover everything:

* 180ee0f - standardize strings a double quoted
* 8ab8808 - mark functions that aren't `process_message` or `timer_event` as local
* 393ecc6  - use `string.format` instead of `..`
* 354f2fa - switch HH logic to https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_security/sandboxes/heka/analysis/moz_security_heavy_hitters.lua
* 31f4d6f9d3b853f5792b31b80d3836acf7639d97 - tabs to spaces
* dca7fa15c06c466831c777af625faac903fdfebe - fix bug from 393ecc6
* c8e307e053edca8ac6cc9e4a9298cc6b09a957a9 - hoist supported hash checks so we can exit early

r? @trink when you get a chance

g-k:
ping @trink

moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
```diff
@@ -0,0 +1,322 @@
...
+```lua
+filename = "moz_security_hh_ip_reputation.lua"
+message_matcher = "Logger == 'input.nginx'"
+ticker_interval = 60
+preserve_data = true
+
+id_field = "Fields[remote_addr]"
+-- id_field_capture = ",? *([^,]+)$"a -- optional e.g. extract the last entry in a comma delimited list
+
+-- hh_items = 10000 -- optional, defaults to 10000 (maximum number of heavy hitter IDs to track), NB: should be <= tigerblood's MAX_ENTRIES setting
```
trink: Docs don't match the default

moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
```diff
@@ -0,0 +1,322 @@
...
+
+-- config for publishing to IP Reputation
+config = {
+   base_url = "https://tigerblood.prod.mozaws.net", -- NB: no trailing slash
+   id = "fxa_heavy_hitters", -- hawk ID
+   key = "8eb54d0469ef000f8819a248c3f4465dc506bb2ad339aaa09466644b2214ffc8", -- hawk secret
```
trink: Name secrets with an underscore `_key` so the are not displayed in the view source UI.

moz_security/sandboxes/heka/output/moz_security_hh_ip_reputation.lua
```diff
@@ -0,0 +1,322 @@
...
+            i = i + 1
+        end
+    end
+
+    -- record multiple for multiple IP, violation pairs
+    -- example body param: [{ip = "127.0.0.1", violation = "fxa:heavy-hitter", weight = 30} ...]
+    -- Note: IPs must be unique
+    local response = json_request("PUT", "/violations/", violations)
+    if 200 <= response.status_code and response.status_code < 300 then
+        return 0
```
trink: timer_event has no return value
http://mozilla-services.github.io/lua_sandbox/heka/output.html

g-k:
Thanks @trink! Ready for another look when you get a chance.

trink:
It looks reasonable.  I don't see it running on any of the logging servers, where was it tested? (or did you want me to package it up and try it out in dev?).

trink:
Looks like I have to investigate a build/packaging error on the openssl module
`/usr/lib/luasandbox/io_modules/openssl.so: symbol X509_check_host, version libcrypto.so.10 not defined in file libcrypto.so.10 with link time reference`

g-k:
Autosquashed and updated the config changes in: https://github.com/mozilla-services/cloudops-deployment/pull/1361

I can send the gpg encrypted hawk secret over too if you want that.

g-k:
Closing in favor of https://github.com/ameihm0912/lua_sandbox_extensions/tree/add-ip-reputation-output-plugin which includes tests and other good stuff.

PR #187, opened 2017-11-30
Changed email to foxsec-dump+OutOfHours

psiinon:


psiinon:
@trink this ok?
I know that @jvehent would like us to stop using foxsec-alerts for auditing type emails like this asap :)

PR #191, opened 2017-12-05
Bugzilla 1423309 - Add the submission and user agent fields

trink:
r? @whd @fbertsch


moz_ingest/io_modules/decoders/moz_ingest/json.lua
```diff
@@ -119,8 +131,18 @@ function transform_message(hsr, msg)
     process_json(hsr, msg)
 
     -- Migrate the original message data after the validation (avoids Field duplication in the error message)
-    msg.Hostname    = hsr:read_message("Hostname")
-    msg.Fields.Host = hsr:read_message("Fields[Host]")
+    msg.Hostname                = hsr:read_message("Hostname")
+    msg.Fields.Host             = hsr:read_message("Fields[Host]")
+    msg.Fields.http_user_agent  = hsr:read_message("Fields[User-Agent]")
```
trink: Should we keep the 'User-Agent' name here?  http_user_agent is what the web server logs use (and where the code was grabbed from)

whd: I don't have a preference.

whd:
LGTM

moz_ingest/io_modules/decoders/moz_ingest/json.lua
fbertsch: No need to keep it if `http_user_agent` is available as well.

trink: I renamed it for the pass through case (so I am not renaming headers)

PR #200, opened 2017-12-20
add lpeg grammar for segfault logging from linux kernel

ameihm0912:
Extends original kernel grammar which handled netfilter logs to also
handle parsing segfault messages.


trink:
Looks goods, please resolve the conflict and squash the commits, thanks.

ameihm0912:
@trink should be good now

Issue #152, opened 2017-07-21
Elasticsearch SSL


Madhu1512:
Is there a plan to add SSL Support for elasticsearch output?

trink:
No, not internally as we don't use ES; I will flag it for community support and see if it gets any interest.

PR #201, opened 2017-12-21
Add SSL support for Elasticsearch output

mikn:
For some reason I cannot identify, I get an error on the setoption passthrough saying that the first parameter isn't of type tcp{any}. I can't really figure out why and I haven't spent a lot of time on it, but if anyone has any ideas, that'd be great. For now, I commented them out, but that's obviously not the state I'll leave it in.
Mainly posting it now to get ideas as to what this is and (potentially) unblock whoever else may have the need for ES SSL support.
Addresses https://github.com/mozilla-services/lua_sandbox_extensions/issues/152


giganteous:
this one can be closed due to #518 being merged

PR #206, opened 2018-01-05
remove definition of CMAKE_INSTALL_PREFIX to zero length string

ameihm0912:
The root CMakeLists.txt defines CMAKE_INSTALL_PREFIX to be a zero length
string, which results in the make install target attempting to install
extensions without any prefix (e.g., /).

Removes definition of CMAKE_INSTALL_PREFIX in CMakeLists.txt so
make install target behaves in the same way as the installation target
in lua_sandbox, which for example defaults to a prefix of /usr/local on
Linux.


trink:
Stripping the prefix is intentional (also this change breaks the unit tests).  During a manual installation DESTDIR is used to put it in the desired location..

ameihm0912:
@trink got it, thanks

Is there a reason the prefix is stripped like that outside of supporting the testing process? It seems like it would be good to have the extensions installation process aligned with lua_sandbox, or would DESTDIR typically be used in both cases and it just happens to not be required in lua_sandbox?

trink:
Yeah, when you use DESTDIR to install to an alternate location you get /DESDTIR/CMAKE_INSTALL_PREFIX/.  Since I manage my manually installed packages with stow the DESTDIR is /usr/local/stow/<package> the additional /usr/local path would have to be manually removed.

PR #207, opened 2018-01-09
update libssl dependencies to allow libssl1.0.2

### LINKED ISSUE #202: ssl module: bad libssl1.0.0 dependency on debian stetch
The ssl module adds a dependency on libssl1.0.0 (>= 1.0.1), see:
https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/ssl/CMakeLists.txt.ssl#L20

But libssl1.0.0 is not available on Debian Stretch (stable) which only has libssl1.0.2 or libssl1.1.

Do you think it could be made compatible with both Debian Jessie and Stretch?

bbinet:
to be compatible with debian stretch and fix #202

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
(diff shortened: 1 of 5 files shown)


trink:
That fix looks good but we should bump the overall cmake requirement to 3.6 and add
CPACK_DEBIAN_PACKAGE_RELEASE 2 to each of the files above

bbinet:
Done in 500a205 (but untested)

PR #208, opened 2018-01-09
remove definition of CMAKE_INSTALL_PREFIX to zero length string

ameihm0912:
The root CMakeLists.txt defines CMAKE_INSTALL_PREFIX to be a zero length
string, which results in the make install target attempting to install
extensions without any prefix (e.g., /).

Removes definition of CMAKE_INSTALL_PREFIX in CMakeLists.txt so
make install target behaves in the same way as the installation target
in lua_sandbox, which for example defaults to a prefix of /usr/local on
Linux.


ameihm0912:
@trink resubmitted version of #206 which should hopefully fix the tests

PR #210, opened 2018-01-10
Update the logfile_roll_by_size to purge old files

cvuillemez:
adds a new setting "roll_retention" to purge old files.
It requires the "lfs" module.

### DIFF
--- a/heka/sandboxes/heka/output/logfile_roll_by_size.lua
+++ b/heka/sandboxes/heka/output/logfile_roll_by_size.lua
@@ -4,6 +4,8 @@
 
 require "io"
 require "string"
+require "os"
+require "lfs"
 
 --[[
 #  Logfile output rolled by size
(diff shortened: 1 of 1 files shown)


trink:
Ideally we want to keep the dependencies between modules to a minimum:
1. A generic log pruning plugin can be introduced in the lfs module
1. This plugin can be moved to the lfs module **if** we keep the lfs dependency
    * the last file_num is preserved on plugin shutdown so it can be used directly instead of computing it with lfs
    * that being said if the plugin is terminated or Hindsight crashes the numbering will start over, the  current behavior would quickly advance to the last file, the new behavior would have those higher number files hanging around until they were reached again (should be very rare but we may want to address it)
1. The pruning operation should be moved to timer_event
1. file_num cannot be set during the initialization as it will be overwritten by the data restoration

cvuillemez:
You're right, but It seems the working of this extension is not usual, cause often when files are rotated:

- older files have the higher suffix number (when a rotation occur, all files are renamed)
- current file has no suffix at all, so it's obvious to write in the right file (even when the app crashe).

If you're agree, I can make a new PR to change it, but it will break the current behavior.

trink:
Please create a new logrotate output plugin in the lfs module with the desired behavior as it is a good counterpart to the tail.lua input.

cvuillemez:
Not sure to well understand. When I speak about lfs module it refers to [this ](https://keplerproject.github.io/luafilesystem/) project (so no pruning/logrotate plugin can be added to it, the project only perform low level functions).
So do you mean:
* create a new output plugin like logfile_roll_by_size.lua but with a different behavior ?
* or create a new plugin to just prune output files (like "prune_input.lua")

trink:
The lfs lua_sandbox_module is here https://github.com/mozilla-services/lua_sandbox_extensions/tree/master/lfs (you will see the tail.lua input I mentioned).  
- If you want a plugin with log rotate functionality that would be a new plugin all together.  
- If the behavior of the existing plugin is acceptable (you just want built in pruning with lfs then it can be migrated to the lua_sandbox_lfs module with the changes)

cvuillemez:
So OK I was confused with the term "LFS" module. I' close this PR and I'll submit a new one.

PR #211, opened 2018-01-11
Issue 209 - Generalize table transformation into a Heka message

trink:


heka/io_modules/decoders/heka/table_to_message.lua
```diff
@@ -0,0 +1,101 @@
...
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+function decode(data, dh)
+    local msg = fn(data)
+    if not msg then return "parse failed" end
+
+    msg = util.table_to_message(msg, cfg.map)
+    -- apply default headers
+    if dh then
+        if not msg.Uuid then msg.Uuid = dh.Uuid end
```
ameihm0912: Are these fields guaranteed to be present here? Should we verify the required fields at all?

trink: All fields are optional. If they aren't set in the message then the default is used, if they aren't set in the defaults then the result is the same the value remains nil.

PR #213, opened 2018-01-12
Add IP reputation analysis and output to Tigerblood

ameihm0912:
This is a continuation of the work from #180 

The existing Tigerblood output functions have been split into a dedicated output module, and a heka module is added to create tigerblood events from the heavy hitters analysis sandbox.


moz_security/sandboxes/heka/output/moz_security_tigerblood.lua
```diff
@@ -0,0 +1,270 @@
...
+
+-- private function for making hawk-signed HTTP requests
+local function json_request(method, uri, body)
+    local body_json = ""
+    if body then
+        body_json = body
+    end
+
+    local headers = {
+        Authorization = hawk_header(
```
trink: Args 2,3,4 can be constructed once outside of this function.  The Arg 3 table can be updated before the call.

moz_security/sandboxes/heka/output/moz_security_tigerblood.lua
```diff
@@ -0,0 +1,270 @@
...
+        client_config["requestor"] = https
+    else
+        client_config["requestor"] = http
+    end
+end
+
+
+-- private function for making hawk-signed HTTP requests
+local function json_request(method, uri, body)
+    local body_json = ""
```
trink:
```lua
if not body then body = "" end
```

rename `body_json` below

trink:
Also, please bump the version on the moz_security package, thanks.

moz_security/sandboxes/heka/output/moz_security_tigerblood.lua
```diff
@@ -0,0 +1,198 @@
...
+
+local function calculate_mac(artifacts)
+    if not supported_hash(client_cfg.algorithm) then
+        error("unsupported hash algorithm")
+    end
+    return mime.b64(hmac.hmac(client_cfg.algorithm, normalize_string(artifacts),
+        client_cfg.key, true))
+end
+
+local function hawk_artifacts(method, resource)
```
trink: These tables (artifacts and options) are created and GC'ed for every request, they can be constructed once and re-used.  These message should be pretty low volume so it is not critical.

moz_security/sandboxes/heka/output/moz_security_tigerblood.lua
```diff
@@ -0,0 +1,198 @@
...
+    end
+    local r, code, resp_headers = client_cfg.requestor.request({
+        method = method,
+        url = string.format("%s%s", client_cfg.base_url, uri),
+        headers = headers,
+        source = ltn12.source.string(body),
+        sink = ltn12.sink.table(response)
+    })
+
+    return {
```
trink: Returned table is not used

ameihm0912: Added a fix to make use of this

PR #214, opened 2018-01-15
io_module to extend heka message with GeoIP information

### LINKED ISSUE #190: Add support for geolocation
eg to allow approx lookup of IP address locations

ameihm0912:
Intended to be used with other io_modules such as decoders; extends
existing message with GeoIP information for any fields that match an
IPv4 address expression. The GeoIP information is added with a postfix
along side the original field value.

Resolves #190

### DIFF
--- a/geoip/CMakeLists.txt
+++ b/geoip/CMakeLists.txt
@@ -9,7 +9,7 @@ externalproject_add(
     ep_geoip
     GIT_REPOSITORY https://github.com/agladysh/lua-geoip.git
     GIT_TAG 06548fd6dac30d4ba0ea5489c3cffe8c20b526c0
-    CMAKE_ARGS ${EP_CMAKE_ARGS}
+    CMAKE_ARGS ${EP_CMAKE_ARGS} -DPARENT_SOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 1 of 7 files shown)


moz_security/io_modules/heka/geo_append.lua
```diff
@@ -0,0 +1,109 @@
...
+    path = "/path/to/geo/dat", -- path to GeoIP data
+    test = false, -- true if being used in tests without GeoIP database
+}
+```
+--]]
+
+local geoip     = require "geoip.city"
+local l         = require "lpeg"
+local lip       = require "lpeg.ip_address"
+local os        = require "os"
```
trink: only the time function is used
```lua
local time = require "os".time
```

moz_security/io_modules/heka/geo_append.lua
```diff
@@ -0,0 +1,109 @@
...
+
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module.
+
+local config = {}
+
+local gtest = {}
+
+-- In test mode, returns Mountain View, US for 192.168.1.2 otherwise nil
+function gtest:query_by_addr(v, lookup)
```
trink: This should be `gtest.query_by_addr`

ameihm0912:
@trink added some fixes based on feedback

Issue #220, opened 2018-01-16
Allow the syslog decoder to cherry pick grammars on a  programname basis


trink:
```lua
decoders_syslog = {
  -- template (string) - The 'template' configuration string from rsyslog.conf
  -- see http://rsyslog-5-8-6-doc.neocities.org/rsyslog_conf_templates.html
  -- Default:
  -- template = "<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag:1:32%%msg:::sp-if-no-1st-sp%%msg%", -- RSYSLOG_TraditionalForwardFormat

  printf_messages = {
   -- array (string and/or array) the order specified here is the load and evaluation order.
     -- string: name of a module containing a `printf_messages` array to import
     -- array: creates an on the fly grammar using a printf format specifications.
       -- see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/modules/lpeg/printf.html

   {"%s:%lu: invalid line", "path", "linenum"},
   "lpeg.openssh_portable", -- must export a `printf_messages` array
  },

  sub_decoders = {
  -- programname_ (string/array)
    -- string: decoder or grammar module name
    -- array: (string and/or array) list of specific messages to parse
      -- string: Sample message used to locate the correct grammar
         -- If no grammar matches the sample message then an error is thrown
         -- and another grammar or module must be added to the printf_messages
         -- configuration. If multiple grammars match the message, the first
         -- grammar with the most specific match is selected.
         -- Note: a special token of `<<DROP>>` and `<<FAIL>>` are reserved for
         -- the last entry in the array to handle the no match case; <<DROP>>
         -- silently discards the message and <<FAIL>> reports an error. If
         -- neither is specified the default no match behavior is to inject the
         -- original message produced by the syslog decoder.
      -- array:
         -- column 1: (string/array)
            -- string: Sample message (see above)
            -- array: printf.build_grammar format specification
         -- column 2: (table/nil)
            -- Transformation table with Heka message field name keys and a
            -- value of the fully qualified transformation function name. The
            -- function returns no values but can error; it receives two
            -- arguments: the Heka message table and the field name to act on.
            -- The function can modify the message in any way.

    nginx  = "decoders.nginx.access", -- decoder module name
    kernel = "lpeg.linux.kernel",     -- grammar module name, must export an lpeg grammar named 'syslog_grammar'
    sshd = {
      -- openssh_portable auth message, imported in printf_messages
[...]

PR #221, opened 2018-01-18
Update the utility function to add geoip info to a Heka message

trink:
Proposed use case:
https://github.com/mozilla-services/lua_sandbox_extensions/issues/220


geoip/io_modules/geoip/heka.lua
```diff
@@ -0,0 +1,112 @@
...
+## Functions
+
+### add_geoip
+
+Given a Heka message add the geoip entries based on the specified field name.
+The requested geoip entries are specified in the `lookup` configuration table.
+The table consists of the entry name and the suffix to add to the field name,
+an empty suffix will overwrite the original value. See: the `opts` variable in
+https://github.com/agladysh/lua-geoip/blob/master/src/city.c for the available
+options, rommon values include "city" and "country_code".
```
ameihm0912: Typo here `rommon -> common`

geoip/io_modules/geoip/heka.lua
```diff
@@ -0,0 +1,112 @@
...
+    if ptime + 3600 < ctime then
+        if geodb then geodb:close() end
+        geodb = assert(geoip.open(cfg.city_db_file))
+        ptime = ctime
+    end
+end
+
+
+local function validate_ip(ip)
+    local sl = #ip
```
ameihm0912: Should we make sure `ip` is a string here?

geoip/io_modules/geoip/heka.lua
```diff
@@ -0,0 +1,112 @@
...
+
+function add_geoip(msg, field_name)
+    if not msg.Fields then return end
+    local value = msg.Fields[field_name]
+    if not value or not validate_ip(value) then return end
+
+    refresh_db()
+
+    for k,v in pairs(cfg.lookup) do
+        local ret = geodb:query_by_addr(value, k)
```
ameihm0912: Should this be wrapped in `pcall`? `query_by_addr` will fail on invalid value for `k`, potentially on invalid addresses

moz_security/tests/hindsight/run/input/generate_sshd.lua
```diff
@@ -30,7 +30,7 @@ function process_message()
         msg.Uuid = v[1]
         msg.Fields.remote_user = v[2]
         msg.Fields.remote_addr = v[3]
-        msg = ga.geo_append(msg)
+        geo.add_geoip(msg, "remote_addr")
```
ameihm0912: I'm guessing even though it's like this in the test, in the implementation we'd have the fields in the sandbox configuration (e.g., `fields = { "remote_addr", "etc" }`)

geoip/io_modules/geoip/heka.lua
trink: No if it is a bad k (configuration error) we should propagate the error)

geoip/io_modules/geoip/heka.lua
trink: Hmm, I am still debating on having this pre-validation at all (let me look at the geoip code and see if we can remove it all together)

moz_security/tests/hindsight/run/input/generate_sshd.lua
trink: No, this cannot be cfg driven (at least not from the geoip_heka cfg) as the field can be variable.  Some logs my have source_ip others remote_addr all handled by the same decoder. See the example in https://github.com/mozilla-services/lua_sandbox_extensions/issues/220

PR #224, opened 2018-01-22
Issue 216 - Add a printf grammar builder

trink:
First issue in the syslog epic branch


lpeg/modules/lpeg/printf.lua
```diff
@@ -0,0 +1,207 @@
...
+
+local sign      = l.S"+-"
+local decimal   = l.P"." * l.digit^0
+local exponent  = l.S"eE" * sign^-1 * l.digit^1
+local float     = sign^-1 * l.digit^1 * decimal^-1 * exponent^-1
+local hdecimal  = l.P"." * l.xdigit^0
+local hexponent = l.S"pP" * sign^-1 * l.digit^1
+local hfloat    = sign^-1 * l.P"0" * l.S"xX" * l.xdigit^1 * hdecimal^-1 * hexponent^-1
+local sp        = l.P" "
+
```
ameihm0912: It doesn't look like string precision to limit output is handled, for example a test case like:

`{{"%.3s", "s"}, "AAABBB", {s = "AAA"}}`

or

`{{"%.3s%.3s", "s", "t"}, "AAABBB", {s = "AAA", t = "BBB"}}`

trink: Fixing: also the first example should actually be a no match

PR #227, opened 2018-01-26
heavy hitters implementation with CMS

ameihm0912:
This is a new implementation of the heavy hitters analysis plugin that attempts to identify high frequency identifiers by calculating a moving average in conjunction with storing frequency per identifier in a CMS data structure.


moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -0,0 +1,154 @@
...
+threshold_cap = 10 -- Threshold will be calculated average + (calculated average * cap)
+-- cms_epsilon = 1 / 10000 -- optional CMS value for epsilon
+-- cms_delta = 0.0001 -- optional CMS value for delta
+```
+--]]
+
+require "string"
+require "table"
+
+local c         = require "streaming_algorithms.cm_sketch"
```
trink: This is only used once so the non-descriptive c. mapping can be removed and simply call
```lua
local cms = streaming_algorithms.cm_sketch.new(cms_epsilon, cms_delta)
```
below

moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -0,0 +1,154 @@
...
+    self.s = {}
+    self.n = 0
+    self.start_time = ostime()
+    self.threshold = 0
+    self.validtick = 0
+    self.evcount = 0
+end
+
+function sampler:calc()
+    if not cms or self.n < sample_min_id or self.evcount < sample_min_ev then
```
trink: cms cannot be nil

moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -0,0 +1,154 @@
...
+
+sampler:reset()
+alist:reset()
+
+function process_message()
+    local id = read_message(id_field)
+    if not id then return -1, "no id_field" end
+    if id_fieldc then
+        id = string.match(id, id_fieldc)
+        if not id then return 1 end -- no error as the capture may intentionally reject entries
```
trink: This should return 0.  1 is fatal and will terminate the plugin

moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -0,0 +1,154 @@
...
+    alist:flush(self.threshold)
+end
+
+function sampler:add(x)
+    if self.start_time + sample_window < ostime() then
+        self:reset()
+        alist:reset()
+        cms:clear()
+    end
+    self.evcount = self.evcount + 1
```
trink: fyi: this count is also in cms:item_count()

PR #230, opened 2018-01-31
Refactor the sub_decoder setup so it can be used in other plugins

trink:
Add an optional mutable argument to decode so sub_decoders know
when it is safe to apply changes directly to the existing message.

### DIFF
--- a/lpeg/io_modules/lpeg/sub_decoder_util.lua
+++ b/lpeg/io_modules/lpeg/sub_decoder_util.lua
@@ -1,0 +1,288 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Sub Decoder Utility Module
+
+Common funtionality to instantiate an LPeg based sub decoder configuration.
+
+## Functions
+
+### load_sub_decoders
+
+Returns a table of sub_decoder functions, keyed by sub_decoder_name.
+
+*Arguments*
+- sub_decoders (table) sub_decoders configuration table
+```lua
+sub_decoders = {
+-- sub_decoder_name (string) = (string/array) a sub_decoder_name of "*" is
+-- treated as the default when the name does not exist.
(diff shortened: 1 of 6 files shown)


moz_logging/io_modules/decoders/moz_logging/line_splitter.lua
```diff
@@ -12,7 +12,8 @@ into multiple messages.
 
 ```lua
 decoders_line_splitter = {
-  sub_decoder = Decoder module name or grammar module name
+  -- printf_messages (table/nil) see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/modules/lpeg/printf.html
+  -- sub_decoder (string/table) see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/io_modules/heka
```
ameihm0912: Is this the right doc link here?

lpeg/io_modules/lpeg/sub_decoder_util.lua
```diff
@@ -0,0 +1,288 @@
...
+- decode (function)
+
+### copy_message
+
+Copies a message for use in decoder/subdecoder
+
+*Arguments*
+* src (table) Heka message table. This is a shallow copy of the individual
+  values in the Fields hash and assumes they will be replaced as opposed to
+  modififed when they are tables. The main use of this function is to populate
```
ameihm0912: modified

trink: I much prefer the single direction flow as the default (it is easier to reason about). You get to operate on the message(s) (each stage could go 1-N) fail , modify, inject and/or pass them down and when inject is called they are captured in the system even if the next phase blows up.

However, the current design does not prevent you from implementing different behavior.   To create the 'root' behavior above, the inject_message function could be replaced in the input plugin so all calls would go through the root's inject_message instance where it could do some finalization (e.g. say you wanted to sign everything).

moz_logging/io_modules/decoders/moz_logging/line_splitter.lua
trink: No... fixing

PR #231, opened 2018-02-01
Dev logrotate

cvuillemez:
Cf [this old PR](https://github.com/mozilla-services/lua_sandbox_extensions/pull/210).
A new output plugin has been added in Heka module (LFS module is not required cause the standard logrotate behavior has no need to compute last filename increment).

### DIFF
--- a/heka/sandboxes/heka/output/logfile_rotate.lua
+++ b/heka/sandboxes/heka/output/logfile_rotate.lua
@@ -1,0 +1,99 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+require "io"
+require "string"
+require "os"
+
+--[[
+#  Logfile output rotated by size
+
+Outputs decoded data stream rotating the log file every time it reaches the
+`rotate_size`.
+
+## Sample Configuration
+```lua
+filename        = "logfile_rotate.lua"
+message_matcher = "TRUE"
+ticker_interval = 0
+
+--location where the payload is written
+output_dir       = "/var/tmp"
+rotate_size      = 1024 * 1024 * 1024
+rotate_retention = 10
+
+-- Specify a module that will encode/convert the Heka message into its output representation.
+encoder_module = "encoders.heka.framed_protobuf" -- default
+```
+--]]
+
+
+local output_dir        = read_config("output_dir") or "/var/tmp"
+local output_prefix     = read_config("Logger")
+local rotate_size       = read_config("rotate_size") or 1e9
+local rotate_retention  = read_config("rotate_retention") or 10
+local encoder_module    = read_config("encoder_module") or "encoders.heka.framed_protobuf"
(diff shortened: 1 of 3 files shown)


trink:
Please rebase this PR against the dev branch and push -f the branch again, thanks.

cvuillemez:
Rebased.

heka/sandboxes/heka/output/logfile_rotate.lua
```diff
@@ -0,0 +1,99 @@
...
+                fn_newer = fn
+            end
+            local fn_older = string.format("%s.%d", fn, file_num)
+
+            -- Apply retention policy (remove older file)
+            if file_num == rotate_retention then
+                local fh_older, err = io.open(fn_older, "r")
+                if fh_older then
+                    fh_older:close()
+                    os.remove(fn_older)
```
trink: You don't really need to check for existence here since a failed remove is being ignored anyway.

heka/sandboxes/heka/output/logfile_rotate.lua
```diff
@@ -0,0 +1,99 @@
...
+                end
+            end
+
+            -- Rename file if exist
+            local fh_newer, err = io.open(fn_newer, "r")
+            if fh_newer then
+                fh_newer:close()
+                local ok, err = os.rename(fn_newer, fn_older)
+                if err then return -1, err end
```
trink: This return code reports an error processing the message but the message was successfully output.  The real problem is that process_message is overloaded (processing messages and performing rotation). The only other option is to move the rotation to timer_event at the cost of more variable sized files, since you can only check the file size a minimum of once a second.  In either case the options for reporting the error are the same error() fatally or ignore it and try again next time (there could be a retry count that would eventually error() out).

heka/sandboxes/heka/output/logfile_rotate.lua
```diff
@@ -0,0 +1,99 @@
...
+            end
+
+            file_num = file_num - 1
```
trink: this could go away with the for loop

heka/sandboxes/heka/output/logfile_rotate.lua
cvuillemez: If you are agree I let the rotation in the process_message, and raise a fatal error when os.rename fail (cause it's an unexpected failure which normally cannot occur).

cvuillemez:
Thanks for your review.
New commit pushed.

heka/sandboxes/heka/output/logfile_rotate.lua
```diff
@@ -0,0 +1,94 @@
...
+
+            local fn_newer = string.format("%s.%d", fn, file_num - 1)
+            if file_num == 1 then -- newer file has never been rotated, so has no suffix
+                fn_newer = fn
+            end
+            local fn_older = string.format("%s.%d", fn, file_num)
+
+            -- Apply retention policy (remove older file)
+            -- (do not perform any check, the file can be missing)
+            if file_num == rotate_retention then
```
trink: The check can be removed because the rename will overwrite it anyway.

PR #236, opened 2018-02-02
hh cms ignore unset ip

ameihm0912:


moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -130,6 +130,9 @@ function process_message()
         id = string.match(id, id_fieldc)
         if not id then return 0 end -- no error as the capture may intentionally reject entries
     end
+    if id == "-" then
```
trink: You can just change the cfg: `id_field_capture = ",? *([^-,]+)$"`

PR #239, opened 2018-02-07
Add Heka message extension to add GeoIP2 data

trink:


trink:
@mreid-moz r? on the Issue 233 commit

mreid-moz:
233 commit looks good to me, with one minor complaint that the `get_geo_city` function is now sort of mis-named, since it returns both city and country.

trink:
Yeah I left it that way as it is the query for the city database

mreid-moz:
Ah, that makes sense. 👍

PR #240, opened 2018-02-08
incorporate hostname value in bastion alert message

ameihm0912:
Refactor bastion alerts and tests. Also incorporates hostname information in bastion alerts along with some other details such as CEP generating host. Adds special handling for delayed reception.


moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
```diff
@@ -5,24 +5,47 @@
...
-Pagerduty will then alert if it is out of policy (hours) for that user.
+Match SSH login events and generate email alerts.
+
+By default, alerts always go to default_email.
+
+If user_email is specified in the configuration, alerts will also be sent to this
+address. User email should be a string containing a %s format specifier which is
+replaced with the username of the account which logged in.
+
+The acceptable_message_delay parameter indicates an age in seconds. If a new event
```
trink: Newer would also be bad

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
```diff
@@ -5,24 +5,47 @@
...
+If delay_email is not set, delayed messages will just be ignored.
+
+This analysis plugin makes assumptions events will be received in a timely manner
+under normal circumstances.
 
 ## Sample Configuration
 ```lua
 filename = "moz_security_sshd_login_monitor.lua"
-message_matcher = "Type == 'logging.shared.bastion.systemd.sshd' && Fields[sshd_authmsg] == 'Accepted'"
+message_matcher = "Type ~= 'logging.*bastion.file.sshd' && Fields[sshd_authmsg] == 'Accepted'"
```
trink: see https://github.com/mozilla-services/cloudops-deployment/pull/1600/files#diff-77b0be4132668975689b7170865893e3R3

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
```diff
@@ -36,24 +59,48 @@ local msg = {
...
+    local ts      = read_message("Timestamp")
+    local hn      = read_message("Hostname") or "unknown"
+    local user    = read_message("Fields[user]") or "unknown"
+    local ip      = read_message("Fields[ssh_remote_ipaddr]") or "unknown"
     local city    = read_message("Fields[ssh_remote_ipaddr_city]")
     local country = read_message("Fields[ssh_remote_ipaddr_country]")
 
-    msg.Fields[2].value    = string.format("%s logged into bastion from %s", user, ip)
-    -- If we also have city and country information, append that to the message
+    if not ts then
```
trink: This check can be removed Timestamp is a require field in a message

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
```diff
@@ -36,24 +59,48 @@ local msg = {
...
+    if user_email then
+        msg.Fields[3].value[2] = string.format(string.format("<%s>", user_email), user)
+    end
+
+    -- finally, if the message is delayed modify the recipient address and add some additional
+    -- information to the payload
+    local delaysec = os.time() - tss
+    if delaysec > acceptable_message_delay then
+        if not delay_email then return -1, "dropping delayed event" end
+        msg.Fields[3].value[1] = string.format("<%s>", delay_email)
```
trink: This turns off the default email forever.

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
```diff
@@ -36,24 +59,48 @@ local msg = {
...
+
+    -- if user_email is set, add a user specific notification address
+    if user_email then
+        msg.Fields[3].value[2] = string.format(string.format("<%s>", user_email), user)
+    end
+
+    -- finally, if the message is delayed modify the recipient address and add some additional
+    -- information to the payload
+    local delaysec = os.time() - tss
+    if delaysec > acceptable_message_delay then
```
trink: We probably want throttling on the delay alerts

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
ameihm0912: Yeah that'd be an odd scenario, likely different root cause; I don't know if we'd want to handle the case of a newer event any differently that the normal case?

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
ameihm0912: They may contain useful information though, by throttle are you thinking drop until we exceed a certain time threshold than create again?

ameihm0912:
@trink fixed primary feedback

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
trink: This is a pretty low volume stream so it is not a must. I was just thinking if it did alert ops will investigate pretty quickly but there is no reason to ping them with more alerts during the investigation window.

moz_security/sandboxes/heka/analysis/moz_security_sshd_login_monitor.lua
trink: I agree on a different root cause but they are both an operational issues.  I would just compare the abs of the delta and alert (with the signed value) if the time appears to be out of wack.

PR #248, opened 2018-02-19
Add an apache log decoder

bplessis:
Extend the lpeg common_log_format to include apache error log format and provide apache log based decoders based on provided LogFormat parser


trink:
I just made a change so we don't have to create decoder wrappers for everything.  It will be merged into master on Feb 26.  I would be a +1 on an Apache error log grammar builder but not a hard coded grammar (unlike the Nginx error log it is user configurable)..  I just created #249, if you are interested in working on it let me know.

Thanks,
Trink

PR #250, opened 2018-02-20
Remove the unsupported call to close() for maxminddb

trink:
This module was ported from geoip and the close() call not removed
causing the reopen and all future lookups to fail.

### DIFF
--- a/maxminddb/io_modules/maxminddb/heka.lua
+++ b/maxminddb/io_modules/maxminddb/heka.lua
@@ -14,9 +14,10 @@ extend the message with GeoIP2 information prior to injection.
 
 Given a Heka message add the geoip entries based on the specified field name.
 The requested geoip entries are specified in the `databases` configuration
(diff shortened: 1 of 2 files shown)


maxminddb/io_modules/maxminddb/heka.lua
```diff
@@ -14,9 +14,10 @@ extend the message with GeoIP2 information prior to injection.
 
 Given a Heka message add the geoip entries based on the specified field name.
 The requested geoip entries are specified in the `databases` configuration
-table. The table is <db_filename> = <lookup table>. The lookup table is
-<suffix> = <path array> (an empty suffix will overwrite the original value). The
-path specification is specific to the database being queried.
+table. The table is &lt;db_filename&gt; = &lt;lookup table&gt;. The lookup table
+is &lt;suffix&gt; = &lt;path arra &gt; (an empty suffix will overwrite the
```
ameihm0912: arra -> array

PR #251, opened 2018-02-21
Issue 246 - Add grammar/function module references to decoders

trink:


lpeg/io_modules/lpeg/sub_decoder_util.lua
```diff
@@ -164,6 +184,20 @@ local function grammar_pick_fn(sd, nomatch_action)
...
+local mod_ref = "([^#]+)#(.+)"
+local function get_module_ref(s)
+    local mname, ref = string.match(s, mod_ref)
+    local mr
+    if mname then
+        mr = require(mname)[ref]
+    else
+        local m = require(s)
+        mr = m.grammar or m.syslog_grammar
+    end
```
ameihm0912: Should we validate mr is valid here? e.g., will an error here be indicated immediately or would it have to wait for the first attempt to use the decoder?

ameihm0912: Actually, disregard on this one -- it looks like we check the result type later on after it's called

PR #253, opened 2018-02-22
Improvements to heavy hitters plugin and supporting code

ameihm0912:
- generalize hawk code
- merge hh functionality into single plugin
- adjust tigerblood output, remove weight parameter
- include travis testing of openssl/ssl
- build time patching for distribution specific openssl


moz_security/sandboxes/heka/analysis/moz_security_hh.lua
```diff
@@ -0,0 +1,187 @@
...
+local violation_type
+local send_tigerblood = read_config("send_tigerblood")
+if send_tigerblood then
+    tbsend = require "heka.tigerblood".send
+    violation_type = read_config("violation_type") or error("violation_type must be configured")
+end
+
+local cms = require "streaming_algorithms.cm_sketch".new(cms_epsilon, cms_delta)
+
+local sampler = {}
```
trink: As discussed the sampler adds a lot of overhead and complexity and should be factored out in future versions.

See logging-cep-test mtrinkala_mozilla.com.hh1.lua for comparison.

moz_security/sandboxes/heka/analysis/moz_security_hh_cms.lua
```diff
@@ -3,7 +3,7 @@
 -- file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 --[[
-#  Mozilla Security Heavy Hitters, CMS + moving average
+# Mozilla Security Heavy Hitters, CMS + moving average
```
trink: This is no longer needed/used as it it replaced by moz_security_hh.lua correct?

moz_security/sandboxes/heka/analysis/moz_security_hh.lua
```diff
@@ -0,0 +1,158 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+#  Mozilla Security Heavy Hitters
+
+See: https://cacm.acm.org/magazines/2009/10/42481-finding-the-frequent-items-in-streams-of-data/abstract
```
trink: Do we still need this reference?

moz_security/sandboxes/heka/analysis/moz_security_hh.lua
```diff
@@ -0,0 +1,158 @@
...
+```lua
+filename = "moz_security_heavy_hitters.lua"
+message_matcher = "Logger == 'input.nginx'"
+ticker_interval = 60
+preserve_data = false
+
+id_field = "Fields[remote_addr]" -- field to use as the identifier
+-- id_field_capture = ",? *([^,]+)$",  -- optional e.g. extract the last entry in a comma delimited list
+
+-- list_max_size = 10000 -- optional, defaults to 10000 (maximum number of heavy hitter IDs to track)
```
trink: The list was only this big for the frequent items calculation (you probably don't want more than 50 or 100 here).

moz_security/sandboxes/heka/analysis/moz_security_hh.lua
ameihm0912: Yeah don't think so, removed

PR #257, opened 2018-03-01
Issue 256 - Update the integration tests to use GeoIP2

trink:
- Fix the X-Forwarded-For split that was removed during refactoring in Issue 233
- Switch to using the last ip in X-Forwarded-For to be consistent with our other analysis


moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -268,7 +260,7 @@ function transform_message(hsr, msg)
         end
     end
 
-    if maxminddb then
+    if maxminddb and not msg.Fields.geoCountry then
```
trink: There is actually a preexisting bug (line 310).  If we are reprocessing records that have already been geocoded and the subdecoder fails the error messages will have two copies of the geo information (from the user perspective it will be invisible as they just fetch the first)

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -281,11 +273,16 @@ function transform_message(hsr, msg)
             hour = current_hour
         end
 
-        local xff = hsr:read_message("Fields[X-Forwarded-For]")
-        local remote_addr = hsr:read_message("Fields[remote_addr]")
-        msg.Fields.geoCity, msg.Fields.geoCountry = get_geo_city(xff, remote_addr)
+        local ip_addr = hsr:read_message("Fields[X-Forwarded-For]")
+        if ip_addr then
+            ip_addr = string.match(ip_addr, "([^, ]+)")
```
whd: It occurs to me now that this is potentially a client-provided string and the only reliable part of the string is the [final ip address](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-for), but since this is the behavior we had before we should probably stick with it.

trink: Yes for the security log we use the last but for telemetry the first has always been used @mreid-moz insight into why?  Or should we actually switch it.

mreid-moz: It was changed from using the entire string to using the first piece in [this commit](https://github.com/mozilla-services/data-pipeline/commit/a29f928e78297433af1c2661cc47572fa704c111) way back in 2015. 

I don't clearly remember the discussion at the time, but I believe that the _last_ entry is the one that ELB actually received the submission from, while the _first_ entry is the one closest to the origin of the request (with the caveat that it may be spoofed). So if a request comes from behind a proxy, the last entry would be the IP of the proxy itself and the first entry would be the earliest IP from before the proxy.

PR #258, opened 2018-03-02
Bug 1174882 - Send an alert when receiving unknown ping types

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1174882)

This sends an alert when an unknown ping type reaches a certain threshold for a day. I've tested this manually on the prod cep.

I can't seem to get the integration tests working properly with the `heka.alert` module. It looks like it uses the wall-clock to set the throttling rate, which does not play well with the fake message stream. I also think the imposed throttling floor of 1 minute gets in the way of testing, but I would also like to verify that the throttling behavior acts properly. 

You might notice in the tests that I send a single message per minute. I'm not sure this is necessary, but I want to verify that the timer event is called at the proper intervals based on the messages that are sent though the queue. I'm relying on the timer events as a clock source to ensure that the running counts are cleared on a daily basis.


moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
```diff
@@ -0,0 +1,140 @@
...
+
+-- Helper function to generate a set
+function Set(list)
+    local set = {}
+    for _, l in ipairs(list) do set[l] = true end
+    return set
+end
+
+-- Maintained in the puppet repository: https://github.com/mozilla-services/puppet-config/blob/master/pipeline/yaml/app/pipeline.yaml#L78
+known_doctypes = Set{
```
trink: You will want to rethink the creation/modification/preservation of this table: 
- To update this table you would have to bump the preservation version (dumping the history) and create a new package every time
- Ideally you would move the definitive list to the cfg and the discovered list would by dynamic and preserved

moz_telemetry/tests/hindsight/run/analysis/new_doctype_monitor.cfg
```diff
@@ -0,0 +1,14 @@
...
+threshold = 10
+
+alert = {
+  disabled = false,
+  prefix = false,
+  threshold = 90,  -- one alert every 90 minutes
+  modules = {
+    email = {recipients = {"amiyaguchi@mozilla.com"}},
+  }
+}
```
trink: Add newline

trink:
Yes anything that calls time() directly will not play well with cli testing (we would need to do something like interpose our own time function to intercept the call and return the current stream time). The other option here would be to create a mock alert module. to test throttling but that doesn't buy you much.

For integration testing you wouldn't want to use the socket alert output (verifying the alert message in the stream like you are doing is sufficient).

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
acmiyaguchi: Updating the list would probably involve moving everything from the discovered list to the known list, which would be worthy of a preservation version bump.

acmiyaguchi:
@trink Ready for another look. I separated the discovered doctypes from known ones, so if there is actually an update to the seeded list it can be updated along with the preservation version.

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
```diff
@@ -0,0 +1,141 @@
...
+local table = require "table"
+
+-- Helper function to generate a set
+function Set(list)
+    local set = {}
+    for _, l in ipairs(list) do set[l] = true end
+    return set
+end
+
+-- Maintained in the puppet repository: https://github.com/mozilla-services/puppet-config/blob/master/pipeline/yaml/app/pipeline.yaml#L78
```
trink: This is another reason this should be cfg based then it only has to be maintained in one place.

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
```diff
@@ -0,0 +1,141 @@
...
+    if refresh_cycles >= seconds_per_day / ticker_interval then
+        refresh_cycles = 0
+        -- make sure to alert in the next cycle if the messages have been throttled
+        for dt, count in pairs(counts) do
+            if count < threshold then counts[dt] = nil end
+        end
+    else
+        refresh_cycles = refresh_cycles + 1
+    end
+end
```
trink: add newline

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
acmiyaguchi: A long list in the configuration feels wrong, but looking at the tests, the configuration can just be a file.

trink: Yes, the cfg will be a file generated by puppet config

acmiyaguchi:
I moved the known doctypes into the configuration.

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_doctype_monitor.lua
```diff
@@ -0,0 +1,109 @@
...
+## Sample Configuration
+```lua
+filename = "moz_telemetry_new_doctype_monitor.lua"
+message_matcher = "Type == 'telemetry'"
+ticker_interval = 60
+preserve_data = true
+threshold = 100
+
+-- Maintained in the puppet repository: https://github.com/mozilla-services/puppet-config/blob/master/pipeline/yaml/app/pipeline.yaml#L78
+known_doctypes = {
```
trink: Heh, you didn't have to include the current full set in the example cfg

PR #266, opened 2018-03-06
Migrate the Hindsight sandboxes into the extensions

trink:
- Reduce duplication of the integration test setup
- Remove circular package dependency


trink:
These need to be moved into sandboxes/heka but have to be removed from Hindsight first to avoid a package installation conflict.

PR #280, opened 2018-03-21
Share some stashed systemd code

renopey:


trink:
This branch needs some work which is why it was stashed.
- convert to use the sub_decoder module
- add integration tests

PR #285, opened 2018-03-26
Bug 1447652 - Add city geoname_id to moz_ingest geoip lookup

relud:


moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -164,18 +164,22 @@ local hour = floor(os.time() / 3600)
 
 local function get_geo_city(ip_addr)
     local city = UNK_GEO
+    local city_geoname_id = UNK_GEO
```
mreid-moz: If this is expected to be an integer, it might be problematic if we default to a string value when failing to look it up. We should probably just leave it out if the lookup fails.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -291,7 +295,7 @@ function transform_message(hsr, msg)
         if not ip_addr then
             ip_addr = hsr:read_message("Fields[remote_addr]")
         end
-        msg.Fields.geoCity, msg.Fields.geoCountry = get_geo_city(ip_addr)
+        msg.Fields.geoCity, msg.Fields.geoCityGeonameId, msg.Fields.geoCountry = get_geo_city(ip_addr)
```
mreid-moz: @trink Do we need to explicitly set the `representation` for the `geoCityGeonameId` field to integer if we don't want it to show up as a float?

moz_ingest/io_modules/decoders/moz_ingest/common.lua
relud: okay, i'll set it to `nil`, and then lua should consider it to not exist when it's put into `msg.Fields`

moz_ingest/io_modules/decoders/moz_ingest/common.lua
trink: The 'value_type' would have to be set to 2 otherwise the numeric value is defaulted to a double.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -164,18 +164,26 @@ local hour = floor(os.time() / 3600)
...
+    if not ok then return city, city_geoname_id, country end
 
     ok, city = pcall(ip.get, ip, "city", "names", "en")
     if not ok then city = UNK_GEO end
 
+    ok, city_geoname_id = pcall(ip.get, ip, "city", "geoname_id")
+    if not ok then
+        city_geoname_id = nil
+    else
+        city_geoname_id = {value = city_geoname_id, value_type = 2}
```
relud: @trink so like this?

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -21,6 +21,11 @@ decoders_moz_ingest_common = {
 
     -- String used to specify GeoIP city database location on disk.
     city_db_file = "/usr/share/geoip/GeoIP2-City.mmdb", -- optional, if not specified no city/country geoip lookup is performed
+    city_report_geoname_id = true, -- optional, if not specified then geoname_id isn't reported
```
relud: > it should become more configuration driven
like this?

moz_ingest/io_modules/decoders/moz_ingest/common.lua
trink: Yes, you can also save a lot of GC if you move the table outside the function and re-use it.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
trink: I was thinking more like the logging inputs. https://mozilla-services.github.io/lua_sandbox_extensions/maxminddb/io_modules/maxminddb/heka.html  but lets just file a different issue for that.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -159,23 +178,48 @@ setfenv(1, M) -- Remove external access to contain everything in the module
...
     ok, city = pcall(ip.get, ip, "city", "names", "en")
     if not ok then city = UNK_GEO end
 
+    if city_report_geoname_id or city_size_set then
+        ok, city_geoname_id = pcall(ip.get, ip, "city", "geoname_id")
+        if not ok then
+            city_geoname_id = nil
+        else
+            GEONAME.value = city_geoname_id
+            city_geoname_id = GEONAME
```
relud: > you can also save a lot of GC if you move the table outside the function and re-use it

like this?

moz_ingest/io_modules/decoders/moz_ingest/common.lua
relud: okay. do i remove this option or leave it in?

trink: Created #287.  We can remove the above unless we don't want the subdivision info in the generic ingestion data in the interim @mreid-moz?

PR #286, opened 2018-03-27
Bug 1447652 - Add subdivisions to moz_ingest geoip lookup

relud:
this may be easier to use, and easier for data review than adding geoname.

### DIFF
--- a/moz_ingest/io_modules/decoders/moz_ingest/common.lua
+++ b/moz_ingest/io_modules/decoders/moz_ingest/common.lua
@@ -21,6 +21,7 @@ decoders_moz_ingest_common = {
 
     -- String used to specify GeoIP city database location on disk.
     city_db_file = "/usr/share/geoip/GeoIP2-City.mmdb", -- optional, if not specified no city/country geoip lookup is performed
+    geo_report_subdivisions = true, -- optional, if not specified then subdivision1 and subdivision2 aren't reported
 
     isp_db_file = "/usr/share/geoip/GeoIP2-ISP.mmdb", -- optional
     isp_docTypes = {customStudy = true}, -- docTypes to perform ISP geoip lookups on, must be set if isp_db_file is defined
@@ -85,6 +86,7 @@ local inject_message       = inject_message
 local sub_decoders  = {}
 local maxminddb
 local city_db
+local geo_report_subdivisions
 local isp_db
 local dedupe
 local duplicateDelta
@@ -119,6 +121,7 @@ local function load_decoder_cfg()
     if cfg.city_db_file then
         maxminddb = require "maxminddb"
         city_db = assert(maxminddb.open(cfg.city_db_file))
+        geo_report_subdivisions = cfg.geo_report_subdivisions
     end
 
     if cfg.isp_db_file then
@@ -164,18 +167,28 @@ local hour = floor(os.time() / 3600)
 
 local function get_geo_city(ip_addr)
     local city = UNK_GEO
+    local subdivision1 = nil
+    local subdivision2 = nil
     local country = UNK_GEO
 
     local ok, ip = pcall(city_db.lookup, city_db, ip_addr)
-    if not ok then return city, country end
+    if not ok then return city, subdivision1, subdivision2, country end
 
     ok, city = pcall(ip.get, ip, "city", "names", "en")
     if not ok then city = UNK_GEO end
 
+    if geo_report_subdivisions then
+        -- subdivisions has at most two items
+        ok, subdivision1 = pcall(ip.get, ip, "subdivisions", 0, "iso_code")
+        if not ok then subdivision1 = nil end
+        ok, subdivision2 = pcall(ip.get, ip, "subdivisions", 1, "iso_code")
+        if not ok then subdivision2 = nil end
+    end
+
     ok, country = pcall(ip.get, ip, "country", "iso_code")
     if not ok then country = UNK_GEO end
 
-    return city, country
+    return city, subdivision1, subdivision2, country
 end
 
 local function get_geo_isp(ip_addr)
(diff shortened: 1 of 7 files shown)


relud:
pgerman expressed concern that geoname isn't a good solution because it puts a "burden to the analyst to determine which portland is portland 29183", so I looked into subdivision because it doesn't require merging on another table for typical use.

I checked GeoLite2-City-Locations-en.csv and the most specific subdivision was able to uniquely differentiate all duplicate cities in the list just as well as including all subdivisions.

```
$ awk -F, '{print $5,$7,$9,$11}' GeoLite2-City-Locations-en.csv | sort -u | wc -l
113016
$ awk -F, '{if($9!=""){print $5,$9,$11}else{print $5,$7,$11}}' GeoLite2-City-Locations-en.csv | sort -u | wc -l
113016
```

alternately we could add `geoSubdivision1` and `geoSubdivision2`, because there is always at most 2 subdivisions specified, based on https://dev.maxmind.com/geoip/geoip2/whats-new-in-geoip2/

r? @mreid-moz

relud:
fixed to use `geoSubdivision1` and `geoSubdivision2`, also I think the subdivisions may only make sense if all are included based on re-reading https://dev.maxmind.com/geoip/geoip2/whats-new-in-geoip2/

> We also provide multiple levels of country subdivision data. The subdivisions these provide correspond to the subdivisions which have been given ISO 3166-2 codes. For example, in the United States, we only provide a single level of subdivision data, corresponding to US states. But for the United Kingdom, we may provide two levels. The first level is the the country (England, Scotland, Wales) or province (Northern Ireland). The second level may be a county, a London borough, a unitary authority, council area, etc.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -291,7 +321,7 @@ function transform_message(hsr, msg)
         if not ip_addr then
             ip_addr = hsr:read_message("Fields[remote_addr]")
         end
-        msg.Fields.geoCity, msg.Fields.geoCountry = get_geo_city(ip_addr)
+        msg.Fields.geoCity, msg.Fields.geoSubdivision1, msg.Fields.geoSubdivision2, msg.Fields.geoCountry = get_geo_city(ip_addr)
```
trink: We should leverage the Field array support here and avoid a breaking data change later.  This allows the use of single field of 'geoSubdivisions'. It should not be an issue for https://bugzilla.mozilla.org/show_bug.cgi?id=1447652#c7 since the immediate use case (US state) only has a single value. Is there a bug to fix the python/scala protobuf readers?

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -164,18 +180,32 @@ local hour = floor(os.time() / 3600)
...
+        -- geoname_id must be in city_size_set
+        local ok, geoname = pcall(ip.get, ip, "city", "geoname_id")
+        if not ok or not city_size_set[geoname] then city = UNK_GEO end
+    end
+
+    -- subdivisions has at most two items
+    ok, subdivision1 = pcall(ip.get, ip, "subdivisions", 0, "iso_code")
+    if not ok then subdivision1 = UNK_GEO end
+    ok, subdivision2 = pcall(ip.get, ip, "subdivisions", 1, "iso_code")
+    if not ok then subdivision2 = UNK_GEO end
```
trink: It seems like unavailable sub divisions should just be nil and the field itself should be optional.

moz_ingest/io_modules/decoders/moz_ingest/common.lua
mreid-moz: I was thinking about the python/scala readers, and I don't see a straightforward way to fix them. The difficulty is that there doesn't seem to be a way to distinguish between a scalar field with one item vs. an array field with one item. As such, we would either have to add special cases for fields we expect to be arrays (gross) or support fields that can be either a single scalar item or an array of items, which would introduce complexity in consumers of the data.

trink: If we change the schema in the future (rename the fields) all down stream consumer would have to change.

trink: Regarding reading the protobuf?  That is pretty standard, so I take it the auto-generated APIs from the protobuf definition are not being used?

mreid-moz: The protobuf definitions are being used, but there is a convenience layer on top of that to allow access to the pings in (roughly) their original JSON structure. This layer includes fields added during decoding in a "metadata" section (similar to how the direct-to-parquet outputs are set up). There's no specification for data types for these fields, so they are all assumed to be scalars (using the first available value of the indicated type).

moz_ingest/io_modules/decoders/moz_ingest/common.lua
relud: made sub divisions default to nil, and made reporting them optional as per https://github.com/mozilla-services/lua_sandbox_extensions/pull/285#pullrequestreview-107684022

moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -267,6 +302,14 @@ function transform_message(hsr, msg)
...
                 city_db = assert(maxminddb.open(cfg.city_db_file))
             end
+            if cfg.city_size_file then
+                city_size_set = {}
+                for line in io.lines(cfg.city_size_file) do
+                    geoname = tonumber(line:match("^%d+"))
+                    assert(geoname ~= nil, "city_size_file lines must all start with integer geoname_id")
+                    city_size_set[geoname] = true
+                end
+            end
```
relud: as requested by @whd

PR #291, opened 2018-04-11
add bit module

ameihm0912:
Add bit operations module

http://bitop.luajit.org/index.html


ameihm0912:
@trink should be fixed up

PR #292, opened 2018-04-12
Add grpc Pub/Sub support

trink:


gcp/tests/hindsight/run/input/load.cfg
```diff
@@ -0,0 +1,25 @@
+filename            = "gcp_pubsub.lua"
```
ameihm0912: Should we just remove all the tests under `tests/hindsight/` for this module since they reference the POC and are not usable outside a pretty specific configuration/environment?

trink: That is why they are under hindsight instead of integration (they are not automatically run but they are extremely useful for the next time we tweak the module).  AWS and Kafka Mirror this but I should set a test configuration of GCP for this.

PR #294, opened 2018-04-13
initial auth_ip_geo analysis sandbox with cidr module

ameihm0912:


moz_security/tests/integration/analysis_auth_ip_geo/run/input/generate_auth.cfg
```diff
@@ -0,0 +1,21 @@
+filename = "generate_auth.lua"
+
+maxminddb_heka = {
+    databases = {
+        ["GeoIP2-City-Test.mmdb"] = {
```
trink: We are starting to get multiple copies of this DB in the repo, should we just start referencing the one here https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/maxminddb/tests/GeoIP2-City-Test.mmdb

moz_security/modules/iputils.lua
```diff
@@ -0,0 +1,203 @@
...
+
+-- Precompute inverted binary subnet masks
+local bin_inverted_masks = {}
+for i=0,32 do
+    local i = tostring(i)
+    bin_inverted_masks[i] = bit.bxor(bin_masks[i], bin_masks["32"])
+end
+
+
+local function split_octets(input)
```
trink: Generally I would use a grammar here (add it to ip_address grammar)
```
local octet =  d8 / tonumber
v4_octet = l.Ct(octet * "." *  octet * "." * octet * "." * octet)
```

moz_security/modules/iputils.lua
```diff
@@ -0,0 +1,203 @@
...
+local string    = require("string")
+
+local M = {}
+setfenv(1, M)
+
+
+-- Precompute binary subnet masks
+local bin_masks = {}
+for i=0,32 do
+    bin_masks[tostring(i)] = bit.lshift((2^i)-1, 32-i)
```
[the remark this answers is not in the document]
ameihm0912: In parse_cidr the mask is accessed through splitting the string (e.g., `"0.0.0.0/0" -> { "0.0.0.0", "0" } -> bin_masks["0"]`). This comes from the code this was based on, we can probably change it if required though.

moz_security/modules/iputils.lua
```diff
@@ -0,0 +1,178 @@
...
+    return bin
+end
+
+
+local function ip2bin(ip)
+    if type(ip) ~= "string" then
+        return nil
+    end
+
+    local octets = split_octets(ip)
```
trink: in-line it?

moz_security/modules/iputils.lua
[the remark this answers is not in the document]
trink: We have access to the integer value everywhere (even after it is parsed out of the string it is tonumber'ed below) so we get rid of a bunch of conversions and tmp variables.  Your call.

moz_security/modules/iputils.lua
```diff
@@ -0,0 +1,178 @@
...
+    if not octets or #octets ~= 4 then
```
trink: Don't need the length check the grammar verified it (the grammar could actually do the complete conversion and this function could go away (optional)).

PR #306, opened 2018-04-25
prototype iomodule for irc integration in output plugins

ameihm0912:
Adds a new IO module that can be used to implement a persistent connection to an IRC server from an output sandbox. The calling Lua sandbox can make use of `new` that will create a background thread, and subsequent calls to `write_chan` on the returned object will queue strings for output in a given channel.

```
i = require "irc"
c = i.new("nick", "server.host", 6697, "#test")
c:write_chan("test message")
```

### DIFF
--- a/moz_security/CMakeLists.txt
+++ b/moz_security/CMakeLists.txt
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
-project(moz-security VERSION 0.0.12 LANGUAGES C)
+project(moz-security VERSION 0.0.13 LANGUAGES C)
 set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Mozilla Infrastructure Security Analysis")
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "${PACKAGE_PREFIX}-streaming-algorithms (>= 0.0.2)")
 string(REGEX REPLACE "[()]" "" CPACK_RPM_PACKAGE_REQUIRES ${CPACK_DEBIAN_PACKAGE_DEPENDS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,7 @@ gcp
 hindsight
 heka
 hyperloglog
+irc
 jose
 kafka
 lfs
--- a/irc/CMakeLists.txt
+++ b/irc/CMakeLists.txt
@@ -1,0 +1,16 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+cmake_minimum_required(VERSION 3.0)
+project(irc VERSION 0.0.1 LANGUAGES C)
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "IRC module")
+set(MODULE_SRCS irc.c)
+set(INSTALL_MODULE_PATH ${INSTALL_IOMODULE_PATH})
+include(sandbox_module)
+
+set(CMAKE_C_FLAGS           "-std=gnu99 -pedantic -Werror -Wall -Wextra -fPIC")
+set(CMAKE_C_FLAGS_RELEASE   "-O2")
+
+find_package(OpenSSL REQUIRED)
+target_link_libraries(irc ${OPENSSL_LIBRARIES})
--- a/irc/index.md
+++ b/irc/index.md
@@ -1,0 +1,85 @@
+# IRC Module
+
+## Overview
+
+The IRC module can be used from Lua to create a persistent background connection to
+an IRC server. The implementation lacks most features, and is intended primarily
+to provide an simple output mechanism for writing real-time operational data from
+a Lua sandbox to an IRC channel.
+
+## Module
+
+### Example Usage
+
+```lua
+i = require "irc"
+conn = i.new("mynick", "server.hostname", 6697, "#mychannel")
+conn:write_chan("Space... the final frontier.")
+```
+
+### Functions
+
+#### new
+
+Create a new persistent background IRC connection. Note that only connections to
+SSL/TLS ports are supported.
+
+```lua
+conn = i.new("mynick", "server.hostname", 6697, "#mychannel")
+```
+
+*Arguments*
+* nick (string)
+* server (string)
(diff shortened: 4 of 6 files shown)


moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  ic->irc_hn = strdup(hn);
+  ic->irc_port = port;
+  ic->irc_chan = strdup(chan);
+  if (chankey != NULL) {
+    ic->irc_chankey = strdup(chankey);
+  }
+  ic->status = STATE_DOWN;
+
+  pthread_mutex_init(&ic->chanqueue_lock, NULL);
+  if (pthread_create(&ic->monitor, NULL, irclib_monitor, ic) != 0) {
```
trink: This thread and mutex should be cleaned up when the user data object is GC'ed.

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  return (1);
+}
+
+static int irclib_writeraw(lua_State *lua)
+{
+  struct ircconn *ic;
+  char buf[1024];
+  const char *s;
+
+  ic = luaL_checkudata(lua, 1, MOZSVC_IRC);
```
trink: You can do the assignment/checks first and fail fast before the 1K buffer is allocated

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  s = luaL_checkstring(lua, 2);
+  snprintf(buf, sizeof(buf), "%s\r\n", s);
```
trink: Do we care if it is truncated?

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  irclib_bio_write(ic->bio, buf, strlen(buf));
+  return (0);
+}
+
+static int irclib_status(lua_State *lua)
+{
+  struct ircconn *ic;
+
+  ic = luaL_checkudata(lua, 1, MOZSVC_IRC);
```
trink: Why separate the declaration from the initialization?

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+static int irclib_parse(struct ircconn *, char *);
+void *irclib_monitor(void *);
+static int irclib_bio_write(BIO *, char *, size_t);
+static int irclib_tryconn(struct ircconn *);
+
+static int irclib_new(lua_State *);
+static int irclib_writeraw(lua_State *);
+static int irclib_writechan(lua_State *);
+static int irclib_status(lua_State *);
+int luaopen_irc(lua_State *);
```
trink: Other modules don't specify the function prototypes unless they are needed before the function definition. Also when provided the argument names are usually included.

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+
+static struct args *irclib_args_parse(char *buf)
+{
+  struct args *ret = NULL;
+  char **tmp;
+  char *t;
+
+  ret = malloc(sizeof(struct args));
+  if (ret == NULL) {
```
trink: Prefer `if (!ret)`

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+    }
+  }
+
+  return (ret);
+}
+
+static void irclib_args_free(struct args *a) {
+  int i;
+
+  for (i = 0; i < a->n; i++) {
```
trink: ++i saves a temporary

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  char **tmp;
+  char *t;
+
+  ret = malloc(sizeof(struct args));
+  if (ret == NULL) {
+    return (NULL);
+  }
+  memset(ret, 0, sizeof(struct args));
+  while ((t = strsep(&buf, " ")) != NULL) {
+    ret->n++;
```
trink: This can be reordered and refactored to use irclib_args_free (instead of having three copies of the cleanup)

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  for (i = 0; i < a->n; i++) {
+    free(a->args[i]);
+  }
+  free(a->args);
+  free(a);
+}
+
+static int irclib_parse(struct ircconn *ic, char *buf)
+{
+  char outbuf[1024];
```
trink: Move to right before it is needed

moz_security/irc.c
```diff
@@ -0,0 +1,550 @@
...
+  ic = luaL_checkudata(lua, 1, MOZSVC_IRC);
+  s = luaL_checkstring(lua, 2);
+  pthread_mutex_lock(&ic->chanqueue_lock);
+  if (ic->chanqueue_size == CHANQUEUE_LENGTH) {
+    pthread_mutex_unlock(&ic->chanqueue_lock);
+    return (0);
+  }
+  q = malloc(sizeof(struct queueent));
+  if (q == NULL) {
+    return luaL_error(lua, "queue allocation failed");
```
trink: long jumps without releasing the mutex

moz_security/irc.c
ameihm0912: I think that's OK yeah, the spec limits the total command length to 512 bytes; the output buffers that would be used from the sandbox are capped at 400 bytes to allow for extra variance in the commands that would prefix the raw text, I will change a few other output locations to enforce a 512 byte cap as well.

moz_security/CMakeLists.txt
```diff
@@ -22,3 +36,8 @@ foreach(test IN LISTS integration_tests)
     CONFIGURATIONS integration
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/integration/${test})
 endforeach()
+
+set(EMPTY_DIR ${CMAKE_BINARY_DIR}/empty)
+file(MAKE_DIRECTORY ${EMPTY_DIR})
+install(DIRECTORY ${EMPTY_DIR}/ DESTINATION ${INSTALL_IOMODULE_PATH} ${DPERMISSION})
+install(TARGETS irc DESTINATION ${INSTALL_IOMODULE_PATH})
```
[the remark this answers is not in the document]
ameihm0912: Seems OK here? I don't see any reason to add a different one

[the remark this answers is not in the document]
trink: Hmm, if I wanted to use an IRC alert I don't think the moz_security module would be the first place I would look.  I think it would make sense to just have a top level IRC module/dir as it is not related to moz_security.

ameihm0912:
r+'d with @trink on IRC

PR #311, opened 2018-05-11
 elasticsearch: don't drop messages when all status are 403

champtar:
When index is read only, all messages will be dropped instead of retrying until
an admin fix the cluster

This is an RFC (not even tested), we might want to only match some status

I'll finish next week, but if you already have comments please share :)

### DIFF
--- a/elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
+++ b/elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
@@ -80,6 +80,8 @@ local req_headers = {
     ["connection"]      = "keep-alive",
 }
 
+local batch_count   = 0
+
 local function send_request() -- hand coded since socket.http doesn't support keep-alive connections
     if not client then client, err = pcreate_client() end
     if err then print(err); return false; end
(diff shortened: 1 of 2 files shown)


elasticsearch/sandboxes/heka/output/elasticsearch_bulk_api.lua
```diff
@@ -107,9 +107,24 @@ local function send_request() -- hand coded since socket.http doesn't support ke
...
+                    local items = doc:find("items")
+                    for _,i in doc:iter(items) do
+                        for _,a in doc:iter(i) do
+                            if (doc:value(doc:find(a, "status")) or 0) >= 400 then
+                                resp_error_count = resp_error_count + 1
+                            end
+                        end
+                    end
+
+                    if resp_error_count >= batch_count then
```
trink: The assumption that if all items failed it must be a cluster problem will most like be incorrect under some conditions so we will inevitably error out when we should have ignored.  There isn't really anything useful in the elastic search API that allows us to easily determine if we should retry or not since there isn't documentation of the various types of errors (we would have to troll the source and manually maintain the set of fatal vs retry errors); which is why the default of log an error and skip was chosen.  I would be a -1 on changing the default behavior without a clear win.

champtar: Would you accept this change if I restrict it to 403 status ?

trink: That would at least address the bad data error vs retry.  What other status codes does it return on a per item bases and under what conditions?

champtar: no idea, and haven't found docs yet :(

champtar: https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/rest/RestStatus.java
https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetaData.java#L139
I would say 401 and 403 could be retriable without side effect (it will not create duplicates or ...)

trink: I would suspect you would not see a 401 at the item level

champtar: yep, just pushed an updated commit

champtar:
Travis build seems broken

champtar:
1) done
2) thanks, I always forgot

PR #313, opened 2018-05-17
add alert module interface to irc io_module

ameihm0912:


irc/io_modules/alert/irc.lua
```diff
@@ -0,0 +1,65 @@
...
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Heka IRC Alert Output Module
+
+## Sample Configuration
+
+```lua
+alert = {
+    modules = {
```
trink: This will only support a single channel I suspect we will want different message to go to different channels (possibly one message to multiple channels).  The channel selection should be part of the alert cfg in irc/modules/heka/alert/irc.lua and this cfg will define all of the acceptable endpoints

irc/modules/heka/alert/irc.lua
```diff
@@ -0,0 +1,30 @@
...
+local cfg = read_config("alert")
+cfg = cfg.modules[module_name]
+assert(type(cfg) == "table", "alert.modules." .. module_name .. " configuration must be a table")
+
+local co = cfg.channel_output
+if type(co) ~= "boolean" then
+    error("channel_output must be boolean")
+end
+
+return {{name = module_name .. ".channel_output", value = co}}
```
trink: Should select the channel here (may want to key on more than channel name, or have a unique name defined in the output cfg)

irc/io_modules/alert/irc.lua
```diff
@@ -0,0 +1,65 @@
...
+
+local irc = require "irc"
+
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+local ircconn = irc.new(cfg.nick, cfg.server, cfg.port, cfg.channel)
+
+function send(msg, mcfg)
+    if not mcfg.channel_output[1] then return nil end
```
trink: This is not an array an should be throwing an error.

irc/io_modules/alert/irc.lua
ameihm0912: The underlying c code only supports a single channel. This was done intentionally to keep the state machine in that code simpler. My thought was that if different output was required, it would just be another instance of the alert module with a different configuration (and a separate connection).

If we want to support channel configuration here, I think we should also look at modifying `irc.c` to support managing multiple channels if you prefer this.

irc/CMakeLists.txt
```diff
@@ -5,12 +5,15 @@
 cmake_minimum_required(VERSION 3.0)
 project(irc VERSION 0.0.1 LANGUAGES C)
```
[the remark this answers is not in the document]
ameihm0912: This hasn't gone into master yet, do we still need a bump?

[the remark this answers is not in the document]
trink: We can leave it as-is then.

PR #316, opened 2018-05-17
add alerting output for tigerblood threshold changes on reputation

ameihm0912:


moz_security/sandboxes/heka/analysis/moz_security_tb_alerts.lua
```diff
@@ -0,0 +1,81 @@
...
+        return 0
+    end
+
+    local t = string.format("[%s] tigerblood adjust %s to %d (set key %s)", prefix, ip, reputation, k)
+    if isviolation then
+        t = t .. string.format(" on violation %s", violation)
+    else
+        t = t .. " directly applied"
+    end
+    alert.send(k, t, t)
```
trink: How many IP addresses are we talking about?  This could get pretty spammy or create a very large alert throttle table.

ameihm0912: Created #317 related to improving the alerting module for this, since we would likely want to do it there vs. have duplicated throttling per sandbox

PR #321, opened 2018-05-23
Bugzilla 1403608 - Allow metadata to be flattened into the root level

trink:


trink:
If you want it backward compatible (schema and data) there are several ways it can be approached, from introducing a new toplevel_metadata group (that is squashed like above), flagging anything with a matching metadata_prefix in the root level schema or updating the schema with an additional optional attribute.  Examples below, I am liking 2 the best
```
    message foo{
        required binary id (UTF8);
        required group metadata {
            required int64  Timestamp;
        }
        required group toplevel_metadata {
             optional binary app_version (UTF8);
        }
```
```
    message foo{
        required binary id (UTF8);
        optional binary <metadata_prefix>app_version (UTF8);
        required group metadata {
            required int64  Timestamp;
        }
```
The metadata attribute support would only be allowed on root level items so it adds complexity without a lot of value.
```
    message foo{
        required binary id (UTF8);
        optional binary app_version (UTF8) [metadata]; # a prefix could be added if desired
        required group metadata {
            required int64  Timestamp;
        }
```

fbertsch:
@trink personally 3 seems the cleanest and most obvious, but 2 works for me as well if the implementation is simpler. Duplicate column names may be a concern here, probably more relevant for 3 if the user doesn't set a prefix.

trink:
Went with the second option:
1. since doesn't break conformance with the parquet-mr spec
1. it will be pretty obvious if duplicate keys exist in the schema as the full names are visible (fyi columns with the same name are allowed in parquet in this case you would just get two copies of the metadata data value)

trink:
@fbertsch yeah that is a different issue (and more general than metadata).  I am assuming you would want the parser to fail it.  We may also want to prevent construction directly with the API; it is a valid schema but pretty useless from a Lua table perspective.

PR #336, opened 2018-06-15
add module for decoding url strings

ameihm0912:


ameihm0912:
@trink this one is updated to use lpeg for parsing

PR #337, opened 2018-06-19
add libinjection analysis sandbox

ameihm0912:


moz_security/sandboxes/heka/analysis/moz_security_webinj.lua
```diff
@@ -0,0 +1,164 @@
...
+-- xss_violation_type = "fxa:webinj_xss" -- required in violations mode, iprepd violation type
+-- sqli_violation_type = "fxa:webinj_sqli" -- required in violations mode, iprepd violation type
+```
+--]]
+
+require "string"
+require "table"
+
+libi    = require "libinjection"
+uri     = require "lpeg.uri"
```
trink: These can/should be local scope

moz_security/sandboxes/heka/analysis/moz_security_webinj.lua
```diff
@@ -0,0 +1,164 @@
...
+    if request_fieldc then
+        req = string.match(req, request_fieldc)
+        if not req then return 0 end -- no error as the capture may intentionally reject entries
+    end
+    if strip_nul then req = string.gsub(req, "%%00", "") end
+
+    local m = uri.uri_reference:match(req)
+    if not m or not m.query then return 0 end
+    local params = uri.url_query:match(m.query)
+    if not params then return 0 end -- don't treat as an error, request may not have parameters
```
trink: fyi: It would only fail to parse if the query string was an "=" or "&"

moz_security/sandboxes/heka/analysis/moz_security_webinj.lua
```diff
@@ -0,0 +1,164 @@
...
+    end
+
+    if foundtype then
+        if not list[id] then
+            list[id] = {
+                cnt         = 0,
+                lastsqli    = nil,
+                lastxss     = nil,
+                sqli        = false,
+                xss         = false
```
trink: The xss and sqli flags seem redundant since you can just the the last* value itself

moz_security/sandboxes/heka/analysis/moz_security_webinj.lua
```diff
@@ -0,0 +1,164 @@
...
+            list_size = list_size + 1
+        end
+        if foundtype == "xss" then
+            list[id].xss = true
+            list[id].lastxss = orig_req
+        else
+            list[id].sqli = true
+            list[id].lastsqli = orig_req
+        end
+        list[id].cnt = list[id].cnt + 1
```
trink: The initial table lookup could be put into a local so you don't have to keep accessing it with the id key.

moz_security/tests/integration/analysis_webinj/run/input/generate_webinj.cfg
```diff
@@ -0,0 +1,22 @@
+filename = "file.lua"
+input_filename = "run/input/access.log"
+
+decoder_module = "decoders.moz_logging.line_splitter"
```
trink: You don't need to use the line splitter (that is only required for the stream kinesis)

PR #340, opened 2018-06-25
add path set frequency analysis sandbox

ameihm0912:


moz_security/sandboxes/heka/analysis/moz_security_pathsetfreq.lua
```diff
@@ -0,0 +1,174 @@
...
+        if v == m.path then
+            foundi = i
+            break
+        end
+    end
+
+    if not w.estab then
+        if foundi == 0 then -- path isn't known
+            if pathsetlen + 1 >= accept_variance then -- addition meets variance requirement
+                w.varied = true
```
trink: We may want to dump the path table when we no longer need it to get it GC'ed a bit quicker (below also).

ameihm0912:
@trink yeah I agree this can be improved; I've fixed the two issues mentioned, I can create an issue regarding the comment -- we can probably resolve this by flushing the established set on variance and recalculating during the existing interval

PR #344, opened 2018-06-28
auth_lastx: process multiple event types in sandbox

ameihm0912:
Adds processing of multiple event types in the auth_lastx analysis
sandbox, so these events can make use of the same serializable tracking
data source.

This also adds a new decoder heka.table_to_fields, that can be used to
process arbitrary nested data (e.g., JSON documents) into a flattened
fields structure for a Heka message.


heka/io_modules/decoders/heka/table_to_fields.lua
```diff
@@ -0,0 +1,102 @@
...
+
+function decode(data, dh)
+    local t = fn(data)
+    if not t then return "parse failed" end
+
+    local f = {}
+    util.table_to_fields(t, f, nil, nil, max_depth)
+    msg = { Fields = f }
+
+    -- apply default headers
```
trink: You could use the simplified version here e.g. https://github.com/mozilla-services/lua_sandbox_extensions/blob/b887cc93674e2493b483cd32a0b44bcd3c0204a8/moz_logging/io_modules/decoders/moz_logging/json_heka_fields.lua#L61

moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
```diff
@@ -49,16 +60,30 @@ default_email = "foxsec-dump+OutOfHours@mozilla.com" -- optional, enable email a
...
--- with the message and will be appended to the alert text as additional informational data
-geocity_field = "Fields[ssh_remote_ipaddr_city"] -- optional, field to extract geo city
-geocountry_field = "Fields[ssh_remote_ipaddr_country"] -- optional, field to extract geo country
+-- expireolderthan = 864000 -- optional, tracked entries older than value are removed, defaults to 864000
+-- lastx = 5 -- optional, track last X entries, defaults to 5
+
+event_fields = {
+    ssh = {
+        select_field     = "Fields[programname]",
+        select_match     = "^sshd$",
```
trink: We could really get rid of this and just perform the match in the message matcher.

moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
```diff
@@ -49,16 +60,30 @@ default_email = "foxsec-dump+OutOfHours@mozilla.com" -- optional, enable email a
...
--- track_fields = { "Fields[ssh_remote_ipaddr_city]", "Fields[ssh_remote_ipaddr_country]" }
-
--- The geocity_field and geocountry_field values are optional, but if set and they are included
--- with the message and will be appended to the alert text as additional informational data
-geocity_field = "Fields[ssh_remote_ipaddr_city"] -- optional, field to extract geo city
-geocountry_field = "Fields[ssh_remote_ipaddr_country"] -- optional, field to extract geo country
+-- expireolderthan = 864000 -- optional, tracked entries older than value are removed, defaults to 864000
+-- lastx = 5 -- optional, track last X entries, defaults to 5
+
+event_fields = {
```
trink: Do we want to make this an array so we can control the order of evaluation?  The key name is this case is basically unused.

moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
ameihm0912: This is just to add an extra layer of precision to make sure we choose the right one in cases where the field names are the same between different events, since we use the key name later on

moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
ameihm0912: The key name is used in the output right now (it's part of the subject to indicate the matched event type)

ameihm0912:
@trink this should be fixed up, I also added a test for the decoder

bbinet:
`heka/io_modules/decoders/heka/table_to_fields.lua` will be also very useful to me, thank you!

That would also be great to have the opposite `fields_to_table.lua` encoder.

heka/io_modules/decoders/heka/table_to_fields.lua
```diff
@@ -0,0 +1,102 @@
...
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+
+function decode(data, dh)
+    local t = fn(data)
+    if not t then return "parse failed" end
+
+    local f = {}
+    util.table_to_fields(t, f, nil, nil, max_depth)
```
trink: separator should be exposed as a cfg option

PR #348, opened 2018-07-06
auth_lastx: support subject field mapping

ameihm0912:


moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
```diff
@@ -161,6 +175,15 @@ function prune_userdata(user, cutoff)
 end
 
 
+function subject_map(f, m)
+    if not m or not f then return f end
+    for k,v in pairs(m) do
+        if k == f then return v end
+    end
+    return f
```
trink: Remove the loop
``` lua
return m[f] or f
```

PR #350, opened 2018-07-12
Adds cloudtrail analysis sandbox

ajvb:
r? @ameihm0912


moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Mozilla Security Cloudtrail notifications
+
```
ameihm0912: Some more details on what the sandbox does would be good here to be consistent with the others

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+## Sample Configuration
+```lua
+filename = "moz_security_cloudtrail.lua"
+message_matcher =
```
ameihm0912: Can we add an example message matcher here, something representative of the actual deployment with the cloudtrail streamer?

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+    for _,v in ipairs(payloadfields) do
+        local vk = string.format("Fields[%s]", v)
+        det[v] = read_message(vk)
+    end
+    return cjson.encode(det)
+end
+
+function get_account_name(account_id)
+    -- { {account_id, account_name}, ... }
+    if cfaccount_name_mapping ~= nil then
```
ameihm0912: `if cfaccount_name_mapping`

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+      for _, v in ipairs(cfaccount_name_mapping) do
+          if v[1] == account_id then return v[2] end
```
ameihm0912: if we change cfaccount_name_mapping to be a hash we can just `return cf_account_name_mapping[k] or account_id` and skip the loop

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+      end
+    end
+    return account_id
+end
+
+function process_message()
+    local event_id = read_message("Fields[eventID]")
```
ameihm0912: minor: to be consistent with other modules `=` aligning some of these early initialization from the message would be good

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+    local account_id = read_message("Fields[recipientAccountId]")
+
+    for _, event in ipairs(cfevents) do
+        local match_counter = 0
+
+        for _, field in ipairs(event['fields']) do
+            local v = read_message(string.format("Fields[%s]", field[1]))
```
ameihm0912: This will result in us reading the same field more than once, for efficiency it might be nice to preload these and iterate over them that way

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+            if v ~= nil then
```
ameihm0912: `if v`

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+                if string.match(v, field[2]) then
+                    match_counter = match_counter + 1
+                end
+            end
+        end
+
+        if match_counter == #event['fields'] then
+              local id = string.format("%s - %s", event["description"], event_id)
+              local s = string.format("%s in %s", event["description"], get_account_name(account_id))
```
ameihm0912: I think this is OK as is for now, it might be nice to have the ability to include information relevant to the event type in the subject (e.g., like a username or other) but we have all that in the payload so maybe not that important for the initial version

moz_security/tests/integration/analysis_cloudtrail/run/analysis/cloudtrail.cfg
```diff
@@ -0,0 +1,33 @@
...
+}
+
+aws_account_mapping = {
+    {"5555555555", "dev"},
+    {"1234567890", "prod"}
+}
+
+alert = {
+    modules = {
+        email = { recipients = { "riker@mozilla.com" } }
```
ameihm0912: `riker`

👌

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,147 @@
...
+--]]
+
+require "cjson"
+
+local alert = require "heka.alert"
+
+local cfevents = read_config("events")
+local cfaccount_name_mapping = read_config("aws_account_mapping")
+
+local payloadfields = {
```
trink: This seems tedious, we could just pull the message down in one shot and not have to request it one field at a time.  However, Fields will be in the native array format http://mozilla-services.github.io/lua_sandbox/heka/message.html

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ajvb: sorry yes! I left this blank because it was not clear to me what the matcher would look like. Is this something we set in a config file?

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ajvb: I wanted to do this, but I was getting parsing errors when there was a number as the key.

both
```
aws_account_mapping = {
    5555555555 = "dev",
    1234567890 = "prod"
}
```
and
```
aws_account_mapping = {
    "5555555555" = "dev",
    "1234567890" = "prod"
}
```
were throwing configuration parsing errors.

Is there documentation on the config file format or is it just a lua file?

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ajvb: :+1: totally agree, but looking at the [`read_message`](https://mozilla-services.github.io/lua_sandbox/heka/analysis.html#read_message) docs, it looks like you cannot do something like `read_message("Fields")`? Am I missing something or is there another way to do this?

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ameihm0912: I think you can specify like:

```
aws_account_mapping = {
    ["5555555555"] = "dev"
}
```

It's just a lua file

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ajvb: fixed

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
trink:
```lua
local msg = decode_message(read_message("raw"))
local j = cjson.encode(msg.Fields) -- however, this will give you the array schema, not the hash
```

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -0,0 +1,160 @@
...
+            { "eventSource", "^iam.amazonaws.com$" },
+            { "recipientAccountId", "^1234567890$" },
+            { "userIdentity.invokedBy", "^signin.amazonaws.com$" },
+            { "userIdentity.sessionContext.attributes.mfaAuthenticated", "^false$" }
+        }
+    }
+}
+
+-- mapping of aws account ids to human-friendly names (optional)
+aws_account_mapping = {
```
trink: Hash would look like this:
```lua
aws_account_mapping = {
    ["5555555555"] = "dev",
    ["1234567890"] = "prod"
}
```

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
ameihm0912: Something like `Type == 'logging.cloudtrail.lambda.cloudtrail.logs|app.log'` should work

It will depend on what we go with for the stream name

ajvb:
ah damn, missed one. fixing + re-squashing

PR #351, opened 2018-07-16
Skip testing

mreid-moz:
Per request from Matt G - messages with `payload.testing == true` should not be considered for alerting.

### DIFF
--- a/moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_experiment_monitor.lua
+++ b/moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_experiment_monitor.lua
@@ -42,6 +42,31 @@ alerted = {
 
 new_experiments = {}
 
+-- Parse submission_text as JSON and return the value of "payload.testing".
+-- If it is not present, return false.
+function is_test(submission_text)
+    if not submission_text then
+        return false
+    end
+
+    local submission = cjson.decode(submission_text)
+    if not submission then
+        return false
+    end
+
+    local payload = submission["payload"]
+    if not payload or type(payload) ~= "table" then
+        return false
+    end
+
+    local testing = payload["testing"]
+    if type(testing) ~= "boolean" then
+        return false
+    end
+
+    return testing
+end
+
 function process_message()
     local ee_txt = read_message("Fields[environment.experiments]")
     if not ee_txt then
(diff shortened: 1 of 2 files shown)


moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_experiment_monitor.lua
```diff
@@ -42,6 +42,31 @@ alerted = {
...
 
+-- Parse submission_text as JSON and return the value of "payload.testing".
+-- If it is not present, return false.
+function is_test(submission_text)
+    if not submission_text then
+        return false
+    end
+
+    local submission = cjson.decode(submission_text)
+    if not submission then
```
trink: The only way line 52 or 77 could fail this test is if null decoding is off and the JSON string is 'null' otherwise the previous line would have thrown an error.  The schema protects against this so it technically is not needed.

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_experiment_monitor.lua
```diff
@@ -42,6 +42,31 @@ alerted = {
...
+    end
+
+    local payload = submission["payload"]
+    if not payload or type(payload) ~= "table" then
+        return false
+    end
+
+    local testing = payload["testing"]
+    if type(testing) ~= "boolean" then
+        return false
```
trink: Should we define 'testing' in the schema so it is not an ad-hoc addition and would prevent some derivative like this "payload": {"testing": 1} in the future.  If so 62-67 can be reduced to `return payload.testing`.

moz_telemetry/sandboxes/heka/analysis/moz_telemetry_new_experiment_monitor.lua
```diff
@@ -42,6 +42,31 @@ alerted = {
...
+    end
+
+    local submission = cjson.decode(submission_text)
+    if not submission then
+        return false
+    end
+
+    local payload = submission["payload"]
+    if not payload or type(payload) ~= "table" then
```
trink: The `not payload` check can be removed the type check is sufficient.

mreid-moz:
Looks like the `payload.testing` field is not actually present in main pings. Closing this since it won't ever match.

PR #355, opened 2018-07-19
Issues 332, 333 - Call Get() before Create()

trink:
The initial version simply called 'create' and ignored an already exists error.
However, that requires more permissions, so now a 'get' is attempted and if
that fails with a not found error then create is tried.

An error is now generated if the subscription topic does not match the specified
topic, this is a safety check to prevent confusion in the case where the subscription
name is already in use on another topic.


bbinet:
This looks good, thank you @trink

PR #359, opened 2018-08-06
Bugzilla 1456610 - Add cross IAM support to the aws.kinesis module

trink:


trink:
@ameihm0912 Let me know if you want Issue 350 broken out; it was a small change that got caught up in this commit.  I could back it out and re-add it as a separate commit to this PR if necessary.

aws/kinesis.cpp
```diff
@@ -233,6 +234,9 @@ get_shard_iterator(simple_consumer *sc, const Aws::String &shardId,
   auto outcome = sc->client->GetShardIterator(sir);
   if (!outcome.IsSuccess()) {
     auto e = outcome.GetError();
+    if (!e.ShouldRetry()) {
+      sequenceId.clear(); // issue 354 invalid checkpoint clear/log and retry
```
ameihm0912: Is there any cases where we wouldn't want to clear on a non-retryable error in `GetShardIterator`?

trink: None that I know of so the only thing we can do here is drop that state, log the error and try again.

PR #360, opened 2018-08-07
add metrics generation for security analysis plugins

ameihm0912:


moz_security/sandboxes/heka/output/moz_security_metrics.lua
```diff
@@ -0,0 +1,118 @@
...
+-- email_recipient = "riker@mozilla.com" -- optional, if set emit metrics to specified recipient
+-- alert = { modules = { email = { ... } } } -- optional, required if email_recipient is set
+```
+--]]
+
+require "cjson"
+
+local debug         = read_config("debug")
+local recipient     = read_config("email_recipient")
+local cephost       = read_config("Hostname") or "unknown"
```
trink: Hostname is always set/available

moz_security/sandboxes/heka/output/moz_security_metrics.lua
```diff
@@ -0,0 +1,118 @@
...
+    if not buf then return 0 end
+    local t = cjson.decode(buf)
+
+    if not z[t.identifier] then
+        z[t.identifier] = {
+            accumulators    = {},
+            uniqitems       = {}
+        }
+    end
+    local x = z[t.identifier]
```
trink: We can make this like the other checks to reduce the number of lookups in the found case.
```
local x = t[y]
if not x then 
   x  = {}
   t[y] = x
end
```

moz_security/modules/heka/secmetrics.lua
```diff
@@ -0,0 +1,142 @@
...
+    self.buf = {
+        identifier      = cfg.identifier,
+        accumulators    = {},
+        uniqitems       = {},
+    }
+end
+
+function secm:send()
+    msg.Fields.metrics = jenc(self.buf)
+    inject_message(msg)
```
trink: If the metrics are empty we should avoid the encoding/injection

moz_security/sandboxes/heka/output/moz_security_metrics.lua
```diff
@@ -0,0 +1,118 @@
...
+        ret[k] = {}
+        for x,y in pairs(v.accumulators) do
+            ret[k][x] = y
+        end
+        for x,y in pairs(v.uniqitems) do
+            local cnt = 0
+            for l,w in pairs(y) do
+                cnt = cnt + 1
+            end
+            ret[k][x] = cnt
```
trink: This could have a collision between accumulator name and a unique item name.

moz_security/sandboxes/heka/output/moz_security_metrics.lua
```diff
@@ -0,0 +1,118 @@
...
+    c = cjson.encode(fold(z))
+    if debug then
+        print(c)
+    else
+        local summary = "secmetrics summary from " .. cephost
+        if malert then -- generate email using email alert module
+            local alertbuf = {
+                Type        = "alert",
+                Payload     = c,
+                Severity    = 1,
```
trink: This is just an informational alert (as the summary is always sent), not actionable correct?  Severity should be 5 (notice) or 6 (informational).

moz_security/sandboxes/heka/analysis/moz_security_auth_lastx.lua
```diff
@@ -295,6 +306,16 @@ function process_message()
         payload = payload .. string.format("WARNING, unacceptable drift %d seconds\n", delaysec)
     end
 
+    if secm then
+        secm:reset()
```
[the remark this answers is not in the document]
ameihm0912: That is true for this particular case, but in cases where we summarize on timer_event (e.g., hh) it would be a larger aggregation. It seems simpler to just generate it in process_event like this since these are low volume vs. have a timer event / interval associated with generating it.

moz_security/modules/heka/secmetrics.lua
```diff
@@ -0,0 +1,144 @@
...
+
+function secm:add_uniqitem(metric, item)
+    self.havedata = true
+    if not self.buf.uniqitems[metric] then
+        self.buf.uniqitems[metric] = {}
+    end
+    self.buf.uniqitems[metric][item] = true
+end
+
+function secm:reset()
```
trink: Reset is still exposed (and used in the plugins)

moz_security/sandboxes/heka/output/moz_security_metrics.lua
```diff
@@ -0,0 +1,120 @@
...
+        ret[k] = {}
+        for x,y in pairs(v.accumulators) do
+            ret[k][x] = y
+        end
+        for x,y in pairs(v.uniqitems) do
+            local cnt = 0
+            for l,w in pairs(y) do
+                cnt = cnt + 1
+            end
+            assert(not ret[k][x], "collision in metric name for " .. k)
```
trink: This will terminate the plugin and end all aggregation, is that the desired behavior?

ameihm0912: Yeah I think this is ok

PR #362, opened 2018-08-08
phabricator access log parsing grammar

ameihm0912:


lpeg/modules/lpeg/phabricator.lua
```diff
@@ -0,0 +1,39 @@
...
+local l     = require "lpeg"
+local pfg   = require "lpeg.printf"
+local d     = require "lpeg.date_time"
+
+local M = {}
+setfenv(1, M)
+
+local tfmt = "[%a, %d %b %Y %H:%M:%S %z]"
+
+fmt = {
```
trink: Should be local

PR #363, opened 2018-08-10
extract auth/usage normalization into submodule for reuse

ameihm0912:


moz_security/modules/heka/selprinc.lua
```diff
@@ -0,0 +1,125 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Heka Select Events with Principal Attributes
+
+This modules provides a generic interface to specify a list a events to match that
```
trink: list of

moz_security/modules/heka/selprinc.lua
```diff
@@ -0,0 +1,125 @@
...
+local ipairs        = ipairs
+local type          = type
+local read_message  = read_message
+local assert        = assert
+
+local M = {}
+setfenv(1, M)
+
+local function cfgchk()
+    assert(type(cfg.events) == "table", "invalid selprinc events configuration")
```
trink: Just assert after reading the cfg var on line 75, you can also get rid of the assert import, or we may want more robust validation here instead of throwing at run time building the return table (as it may not fail for some time)?

PR #366, opened 2018-08-10
Add attribute support for GCP pub/sub

trink:
- Move batching from the plugin into the library

### DIFF
--- a/gcp/common.cpp
+++ b/gcp/common.cpp
@@ -1,0 +1,164 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
+/* vim: set ts=2 et sw=2 tw=80: */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/** @brief GCP common functions  @file */
+
+#include "common.h"
+
+#include <climits>
+#include <iomanip>
+#include <sstream>
+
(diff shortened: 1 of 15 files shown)


gcp/common.cpp
```diff
@@ -0,0 +1,161 @@
...
+    case LSB_PB_STRING:
+      {
+        int tag = 0;
+        int wiretype = 0;
+        while (p && p < e) {
+          p = lsb_pb_read_key(p, &tag, &wiretype);
+          if (wiretype == LSB_PB_WT_LENGTH) {
+            long long len;
+            p = lsb_pb_read_varint(p, e, &len);
+            if (p && (len >= 0 || p + len <= e)) {
```
ameihm0912: is it possible here for `p + len` to overflow negative?

gcp/sandboxes/heka/input/gcp_pubsub.lua
```diff
@@ -65,46 +68,91 @@ local err_msg = {
...
-                        err_msg.Payload = err
-                        err_msg.Fields.data = data
-                        pcall(inject_message, err_msg)
-                    end
+
+                local ok, err = pcall(decode, msgs[i][1], msg, true)
+                if not ok or err then
+                    err_msg.Payload = err
+                    err_msg.Fields.data = data
+                    pcall(inject_message, err_msg)
```
ameihm0912: does any failure here need to be handled?

trink: no, if we fail injecting the failure, we simply drop it

gcp/common.cpp
trink: yes

PR #370, opened 2018-08-17
Support multiple matchers per field for cloudtrail

ajvb:
This is particularly useful for when matching against multiple aws account ids. As well, in moz_security_cloudtrail we are currently using `string.match` which does not have an `or` operator.

As well, this PR adds an example of a `not` operation within `string.match`.

### DIFF
--- a/moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
+++ b/moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
@@ -48,7 +48,7 @@ events = {
             { "eventSource", "^iam.amazonaws.com$" },
             { "recipientAccountId", "^1234567890$", "^1122954321$" },
             { "userIdentity.invokedBy", "^signin.amazonaws.com$" },
-            { "userIdentity.sessionContext.attributes.mfaAuthenticated", "[^true]" }
+            { "userIdentity.sessionContext.attributes.mfaAuthenticated", "^false$" }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -119,9 +119,14 @@ function process_message()
             end
 
             if det[field[1]] then
-                if string.match(det[field[1]], field[2]) then
+              -- All items in the event field after the first one are potential matches
+              local i = 2
+              while field[i] do
+                if string.match(det[field[1]], field[i]) then
                     match_counter = match_counter + 1
```
trink: Should break after the first successful match (saves some processing and prevents match_counter from potentially being off if multiple items matched)

moz_security/sandboxes/heka/analysis/moz_security_cloudtrail.lua
```diff
@@ -46,9 +46,9 @@ events = {
         description = "IAM action in production account from console without mfa",
         fields = {
             { "eventSource", "^iam.amazonaws.com$" },
-            { "recipientAccountId", "^1234567890$" },
+            { "recipientAccountId", "^1234567890$", "^1122954321$" },
             { "userIdentity.invokedBy", "^signin.amazonaws.com$" },
-            { "userIdentity.sessionContext.attributes.mfaAuthenticated", "^false$" }
+            { "userIdentity.sessionContext.attributes.mfaAuthenticated", "[^true]" }
```
trink: The negated set is a misleading/bad configuration example and is a hack that wouldn't work if you wanted to negate something like recipientAccountId.

ajvb: That is really good to know, thanks. Will remove for this PR and look into switching to lpeg.re.

PR #371, opened 2018-08-21
Use lpeg.re for matching in cloudtrail sandbox

ajvb:
Switch to using lpeg.re within cloudtrail analysis sandbox so that field
matchers can be more expressive. Also allows for the removal of logic
supporting multiple matchers against an event field.

r? @ameihm0912 @trink


trink:
A quick test showed the optimization running ~1.5x faster (updating the for loop and caching the Field string generation)

PR #380, opened 2018-08-31
patch openssl, only allocate/deallocate locks if unset or local

ameihm0912:


trink:
This looks like something we should upstream.

Issue #381, opened 2018-09-05
Standard decoders should support transformations


ameihm0912:
Looks like with subdecoder util etc transformations are not possible, and they only work on grammars/printf/etc

It would be good if this could be extended to work with something like decoders.heka.table_to_fields

trink:
Technically it is doable.  We have to trap all inject_message calls, insert the transformation in the wrapper and then call the real inject message. 

**Caveat**: this will only work with decoders calling inject_message with a table anything else would throw an error at this point as the APIs to support the transformation in the Heka Stream Reader or protobuf string are not exposed.

ameihm0912:
https://github.com/mozilla-services/lua_sandbox_extensions/pull/388

PR #382, opened 2018-09-05
Extract usernames from Taskcluster audit logs

ajvb:
Adds the `taskcluster#add_username` func that can be used with another
decoder to extract the ldap username from a taskcluster clientid. This
makes plugging into other analysis sandboxes much simpler.

Blocked by https://github.com/mozilla-services/lua_sandbox_extensions/issues/381
r? @ameihm0912

### DIFF
--- a/moz_security/io_modules/taskcluster.lua
+++ b/moz_security/io_modules/taskcluster.lua
@@ -59,7 +59,7 @@ function add_username(msg, field_name)
     for _, matcher in ipairs(cfg.matchers) do
       username = re.match(clientid, matcher)
       if username then
-        msg.Fields["username"] = username
+        msg.Fields[username_field] = username
         break
       end
     end
--- a/moz_security/io_modules/taskcluster.lua
+++ b/moz_security/io_modules/taskcluster.lua
@@ -33,7 +33,7 @@ taskcluster = {
 }
 decoder_module = {
     {
-        { "decoders.heka.table_to_fields" },
+        "decoders.heka.table_to_fields",
         { clientid = "taskcluster#add_username"}
     }
 }
--- a/moz_security/io_modules/taskcluster.lua
+++ b/moz_security/io_modules/taskcluster.lua
@@ -1,0 +1,51 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+# Heka Message Extensions for Taskcluster
+
+This module is intended to be used with another IO module such as a decoder to
+add a username field if the client is a human user.
+
+## Functions
+
+### add_username
+
+Given a Heka message with a Taskcluster clientid, attempt to parse out the username
+associated with the clientid. This is currently specific to Mozilla LDAP usernames,
+as that is the primary way in which human users authenticate with Taskcluster. If no
+username is found, do nothing.
+
+*Arguments*
+- msg (table) - original message
+- field_name (string) - field name in the message to lookup
+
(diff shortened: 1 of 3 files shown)


moz_security/io_modules/taskcluster.lua
```diff
@@ -0,0 +1,51 @@
...
+local re = require "re"
+
+local M = {}
+setfenv(1, M)
+
+function add_username(msg, field_name)
+    if not msg.Fields then return end
+    local clientid = msg.Fields[field_name]
+
+    username = re.match(clientid, "'mozilla-auth0/ad|Mozilla-LDAP|' (s <- {%w+}) '/'")
```
ameihm0912: should we make the auth0/LDAP/etc part of this a config option so it's not hardcoded?

moz_security/io_modules/taskcluster.lua
```diff
@@ -0,0 +1,51 @@
...
+    if username then
+      msg.Fields["username"] = username
```
ameihm0912: We might want to do something like we do with maxmind so the field is configurable

https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/maxminddb/io_modules/maxminddb/heka.lua#L34

moz_security/io_modules/taskcluster.lua
```diff
@@ -0,0 +1,68 @@
...
+## Configuration examples
+taskcluster_decoder = {
+    username_field = "username",
+    matchers = {
+      "'mozilla-auth0/ad|Mozilla-LDAP|' (s <- {%w+}) '/'"
+    }
+}
+decoder_module = {
+    {
+        { "decoders.heka.table_to_fields" },
```
trink: This proposed syntax is ambiguous and conflicts with the grammar builder syntax.

ajvb: Agreed. Going to wait till https://github.com/mozilla-services/lua_sandbox_extensions/issues/386 is fleshed out as this is taken from the syntax that @ameihm0912 was using before creating #381

moz_security/io_modules/taskcluster.lua
```diff
@@ -0,0 +1,68 @@
...
+setfenv(1, M)
+
+function add_username(msg, field_name)
+    if not msg.Fields then return end
+    local clientid = msg.Fields[field_name]
+
+    for _, matcher in ipairs(cfg.matchers) do
+      username = re.match(clientid, matcher)
+      if username then
+        msg.Fields["username"] = username
```
ameihm0912: should this be `username_field`?

moz_security/io_modules/taskcluster.lua
```diff
@@ -0,0 +1,68 @@
...
+
+*Arguments*
+- msg (table) - original message
+- field_name (string) - field name in the message to lookup
+
+*Return*
+- none - the message is modified in place if a username is found
+
+## Configuration examples
+taskcluster = {
```
ameihm0912: i think this needs to be updated for the module cfg / module name change

ajvb: do you mean the change to having `local module_name = ...`? If so, no this is correct. Originally it was `taskcluster_decoder` with `local module_name = "taskcluster_decoder"`

ameihm0912: right, sorry yeah -- disregard this one

moz_security/io_modules/taskcluster.lua
trink: 381 is in (the first argument is a sting since decoders don't take any additional options)
```lua
decoder_module = { {"decoders.heka.table_to_fields", {clientid = "taskcluster#add_username"}} }
```

PR #383, opened 2018-09-06
support lookup modules in alert, add idrouter, auth_lastx lookup

ameihm0912:


ajvb:
Something that I thought about while reading through this is that in the near future we will probably want to support teams as well. Now, we may be able to just use the `user` functionality here for that, as we may not be mixing team-alerting and user-alerting, but it's something that we may want to consider.

An example where they could be mixed is within cloudtrail alerting:

1. An alert is trigger for the removal of MFA on `picard`. We will want to alert `picard` then escalate to `security`.
2. An alert is trigger because a cloudtrail is turned off. We may want to alert the team's pagerduty.


I don't think this needs to be addressed in this PR, but wanted to mention this as I believe it will be functionality we will want soon.

ameihm0912:
@ajvb thanks for the feedback, yeah see what you are saying in your comment for sure, there is a layer between users and global missing here. I feel like this could evolve more but provides a start at least, my hope here was to get something more flexible in place so we can avoid manually creating alert events when we need dynamic alerting without breaking the existing alert API.

I think the stuff you mentioned we can probably adjust the code for when we need it and evolve things, do it via CI, or do it externally too like you mentioned.

moz_security/modules/heka/alert/idrouter.lua
```diff
@@ -0,0 +1,183 @@
...
+be used of senduser is set to true in lookup data.
+
+Global notification is used when sendglobal is set to true in the lookup data.
+
+Error notification is used when senderror is set to true in the lookup data.
+
+A catchall setting can be set for each category for both IRC and email alerts. If the subject
+entry does not specify a given category setting, the catchall will be used for the alert.
+
+If lookup data is specified that does not contain a subject entry or does not match any known
```
trink: It seems a bit tedious to have to constantly manage the direct alert user list and to specify a map when X = X.  How will this work with our current lastx monitor where we send a direct alert for all users?

moz_security/modules/heka/alert/idrouter.lua
```diff
@@ -0,0 +1,183 @@
...
+a table containing configuration parameters.
+
+At the least, each subjects entry should contain a mapfrom parameter. This is used to locate
+the correct subject entry using the lookup data subject string. Where the lookup data subject
+string matches an entry in mapfrom, this identity configuration will be used for the alert.
+
+email and IRC alerts are currently supported. For each time, three categories exist.
+
+Direct notification is used to determine how to route the alert directly to the user, and will
+be used of senduser is set to true in lookup data.
```
trink: /of/if/

moz_security/modules/heka/alert/idrouter.lua
```diff
@@ -0,0 +1,183 @@
...
+
+The subjects configuration contains any user specific alerting parameters. Each entry
+within subjects should be a key, which represents the standardized user identity string, and
+a table containing configuration parameters.
+
+At the least, each subjects entry should contain a mapfrom parameter. This is used to locate
+the correct subject entry using the lookup data subject string. Where the lookup data subject
+string matches an entry in mapfrom, this identity configuration will be used for the alert.
+
+email and IRC alerts are currently supported. For each time, three categories exist.
```
trink: phrasing

moz_security/modules/heka/alert/idrouter.lua
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    local eur
+    local eer
+    local egr
+    local iut
+    local igt
+    local iet
+
+    local u, udata = lalias(ldata.subject)
```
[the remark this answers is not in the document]
ameihm0912: Yeah I think erroring here is fine, that would mean something isn't using the module correctly

ajvb:
@ameihm0912 :+1: agreed.

PR #395, opened 2018-09-27
Issue 282 - Add a GCP logging (StackDriver) module

trink:


gcp/logging.cpp
```diff
@@ -0,0 +1,642 @@
...
+    return luaL_error(lua, "invalid " LSB_THIS_PTR);
+  }
+  ww->w->logger = lsb_get_logger(lsb);
+#endif
+  luaL_getmetatable(lua, mt_writer);
+  lua_setmetatable(lua, -2);
+
+  bool err = false;
+  try {
+    auto creds = grpc::GoogleDefaultCredentials();
```
ameihm0912: Is this instance default? Should this support service account JSON files and the like?

trink: It works for both: https://grpc.io/docs/guides/auth.html#using-google-token-based-authentication

PR #398, opened 2018-10-11
Sprint 18

trink:


moz_telemetry/modules/moz_telemetry/histogram.lua
```diff
@@ -210,6 +210,7 @@ local function debug_histogram_graph(h, pcc, row, closest, title, graphs)
     local curr = h.data:get_row(row)
     local cdata = {}
     for x,y in ipairs(curr) do
+        if y ~= y then y = 0 end
```
ameihm0912: is `y ~= y` what we want here? I'm not sure what this is doing

Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN yum makecache && \
     postgresql-devel libcurl-devel lua-devel && \
     yum install -y https://packages.red-data-tools.org/centos/red-data-tools-release-1.0.0-1.noarch.rpm && \
     yum install -y libmaxminddb-devel jq libmaxminddb librdkafka-devel jansson-devel devtoolset-6 && \
-    yum install -y --enablerepo=epel arrow-devel-0.9.0 && \
+    yum install -y --enablerepo=epel --nogpgcheck arrow-devel-0.9.0 && \
```
ameihm0912: is there a way we could avoid removing the signature check like loading a key for this somewhere? or is the package just not signed?

trink:
```
warning: /var/cache/yum/x86_64/7/red-data-tools/packages/arrow-devel-0.9.0-1.el7.centos.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID 079f8007: NOKEY
Public key for arrow-devel-0.9.0-1.el7.centos.x86_64.rpm is not installed
```

That was just the quickest fix, looking to see if there is an easy way to load the key here.

PR #409, opened 2018-11-21
 Issue 408 - Default the GRPC send/receive message size to unlimited 

trink:


gcp/pubsub.cpp
```diff
@@ -148,7 +148,13 @@ static int publisher_new(lua_State *lua)
   bool err = false;
   try {
     auto creds = grpc::GoogleDefaultCredentials();
-    pw->p->stub = std::make_unique<Publisher::Stub>(grpc::CreateChannel(channel, creds));
+    // see https://github.com/googleapis/google-cloud-node/pull/2007
+    // Fix the send/receive size mis-match in
+    // https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h#L396
+    grpc::ChannelArguments cargs;
+    cargs.SetMaxSendMessageSize(-1); // make sure this remains -1 even if the todo is completed
+    cargs.SetInt(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS, 5000);
```
whd: Was this the default or empirically determined to be a good value?

gcp/pubsub.md
```diff
@@ -63,7 +63,10 @@ local msgs, cnt = subscriber:pull(batch_size)
 * batch_size (integer) Number of items in a single request (1-1000)
 
 *Returns*
-* msgs (array/nil) One or more more message (can throw on error)
+* msgs (array/nil) One or more messages (can throw on error)
+    The array contains and array for each message, column one is the data
```
whd: "the array contains an array"

gcp/pubsub.cpp
trink: We haven't actually had an issue with this so we use the same guesstimate as the node library.

gcp/pubsub.md
trink: fixed

PR #415, opened 2019-01-23
Bug 1339024 - Protect against dimension file names greater than 255

trink:


whd:
For whatever reason, sometimes AWS gives the hostname a fully qualified internal domain e.g. +1542909993_32_ip-172-31-29-209.us-west-2.compute.internal.done. We might want to strip the extra bits off the end or increase the suffix allowance from 40 to 64.

Additionally we discovered that `buffer_cnt` is higher than we were expecting (many rename operations occur within the same second), and could potentially reach 3 digits.

Otherwise LGTM.

PR #420, opened 2019-03-11
Bug 1533126 - Add an option in moz_ingest to keep the IP address

trink:


moz_ingest/io_modules/decoders/moz_ingest/common.lua
```diff
@@ -22,6 +22,7 @@ decoders_moz_ingest_common = {
     -- String used to specify GeoIP city database location on disk.
     city_db_file = "/usr/share/geoip/GeoIP2-City.mmdb", -- optional, if not specified no city/country geoip lookup is performed
     geo_report_subdivisions = true, -- optional, if not specified then subdivision1 and subdivision2 aren't reported
+    -- geo_ip_keep = false, -- optional, when specified the IP address is saved in the `geoIP` field
```
whd: Any reason to not simply reuse (i.e. not remove) the field that the ip address originally comes from? This is perhaps related to multiple potential sources for geoip information.

trink: The original message and the new message are passed to the subdecoder.  The new message (the one the telemetry decoder will ultimately end up injecting does not have the data (so we add it here).  Technically the telemetry decoder could pull it from the original message by repeating the x-forwarded-for, remote_addr test/parsing but it just seem cleaner to do it here.

PR #429, opened 2019-04-08
Bug 1542012 - Don't restrict telemetry docType monitors to the Firefo…

trink:
…x application


trink:
@whd
timer_event() run/analysis/moz_ingest_doctype_monitor.lua:463: instruction_limit exceeded on the sync ping (that would be 1MM)  restarted and the current max is 2374.  Not sure if this was caused by lifting the restriction.  Looking at the code I believe the only thing that could blow it out this was is a huge list of ingestion error diagnostics.

PR #438, opened 2019-06-06
Add an AMQP module

trink:
- Consumer only, needed to read the CI jobs exchange


bbinet:
Thanks, this could be a good alternative to heka_tcp plugin.
Do you plan to implement AMQP producer too?

trink:
It is low priority but eventually yes.

PR #441, opened 2019-06-17
Implement payload_keep option for subdecoders

giganteous:
Hello Trink,

This patch is written to solve my problem, which was that the original Payload dropped off of my messages when using a sub_decoder in the syslog module.

I'm just hacking away, so do tell me if I need to proceed in a different way. I kept an eye on how the nginx module supports this option, and added a small integration test.

### DIFF
--- a/syslog/io_modules/decoders/syslog.lua
+++ b/syslog/io_modules/decoders/syslog.lua
@@ -13,7 +13,7 @@ decoders_syslog = {
   -- see https://github.com/rsyslog/rsyslog/blob/de0a9dd10703c4e6efcc69164781220d31a9e115/runtime/rsconf.c#L85
   -- Default:
   -- template = "<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag:1:32%%msg:::sp-if-no-1st-sp%%msg%", -- RSYSLOG_TraditionalForwardFormat
-  -- payload_keep = false, -- Flag indicating if the original payload should be kept when using sub_decoders
+  -- keep_data = nil, -- Flag indicating if the complete input in the returned data, and how to name that field.
 
   -- printf_messages = nil, -- see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/modules/lpeg/
   -- sub_decoders = nil, -- see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/io_modules/lpeg/sub_decoder_util.html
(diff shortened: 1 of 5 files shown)


syslog/io_modules/decoders/syslog.lua
```diff
@@ -105,6 +106,8 @@ function decode(data, dh, mutable)
 
     local df = sub_decoders[programname]
     if df then
+        if cfg.payload_keep then msg.Payload = data end
```
trink: Not sure I like the inconsistency in the msg.Payload between the df and non df case (generally we have just been storing the message part in Payload (the non df and printf cases)).  This would also be sub_decoder dependent as it can overwrite or erase it.  If you need the full message we may be better off renaming it (cfg and destination field) and allowing it to work for the df and non df cases.  Thoughts?

giganteous: I agree. 

It seems this code is passing my test:
```diff
@@ -13,7 +13,7 @@ decoders_syslog = {
   -- see https://github.com/rsyslog/rsyslog/blob/de0a9dd10703c4e6efcc69164781220d31a9e115/runtime/rsconf.c#L85
   -- Default:
   -- template = "<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag:1:32%%msg:::sp-if-no-1st-sp%%msg%", -- RSYSLOG_TraditionalForwardFormat
-  -- payload_keep = false, -- Flag indicating if the original payload should be kept when using sub_decoders
+  -- keep_data = nil, -- Flag indicating if the complete input in the returned data, and how to name that field.
 
   -- printf_messages = nil, -- see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/modules/lpeg/
   -- sub_decoders = nil, -- see: https://mozilla-services.github.io/lua_sandbox_extensions/lpeg/io_modules/lpeg/sub_decoder_util.html
@@ -54,6 +54,7 @@ local sdu           = require "lpeg.sub_decoder_util"
 
 local cfg = read_config(module_cfg) or {}
 assert(type(cfg) == "table", module_cfg .. " must be a table")
+assert(type(cfg.keep_data) == "string" or type(cfg.keep_data) == nil, "keep_data must be nil, or a string: " .. cfg.keep_data)
 local template      = cfg.template or "<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag:1:32%%msg:::sp-if-no-1st-sp%%msg%"
 local grammar       = syslog.build_rsyslog_grammar(template)
 local sub_decoders  = sdu.load_sub_decoders(cfg.sub_decoders, cfg.printf_messages)
@@ -104,10 +105,13 @@ function decode(data, dh, mutable)
     fields.msg = nil
     sdu.add_fields(msg, fields)
 
+    if cfg.keep_data ~= nil then
+        sdu.add_fields(msg, { [cfg.keep_data] = data })
+    end
+
+
     local df = sub_decoders[programname]
     if df then
-        if cfg.payload_keep then msg.Payload = data end
-
         local err = df(payload, msg, true)
         if err then
             err = string.format("%s.%s %s", module_name, programname, err)

```

Are these ``if cfg.keep_data ~= nil`` good enough? I'm still in my teens wrt writing lua you see.

syslog/io_modules/decoders/syslog.lua
```diff
@@ -103,6 +107,11 @@ function decode(data, dh, mutable)
     fields.msg = nil
     sdu.add_fields(msg, fields)
 
+    if cfg.keep_data ~= nil then
```
trink: Move this to after line 104 and it can be simplified like the following
```lua
if cfg.keep_data then fields[cfg.keep_data] = data end
```

giganteous: Cheers. Will do

PR #442, opened 2019-07-03
Sendmail

giganteous:
Grammar for sendmail logs.

Using the keyvalue grammar in combination with other grammar was a weird experience, but managed it.


syslog/modules/lpeg/sendmail.lua
```diff
@@ -0,0 +1,32 @@
...
+l.locale(l) -- we want l.alpha etc...
+
+local rawset = rawset
+
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+local space = l.space^0
+local key = l.C(l.alpha^1) * space
+local value = l.C((l.P(1) - l.P', ')^1) * space
```
trink: space can be removed (it can never match). 

Based on the spec lines (21-26) could be simplified down to the following:
```lua
local sep   = l.P', '
local key   = l.C(l.alpha^1)
local value = l.C((l.P(1) - sep)^1)
local pair  = l.Cg(key * '=' * value) * sep^-1
```
If you want to be more cautious about extra white space 
```lua
local sep = l.P',' * l.space^1
local key = l.C(l.alpha^1) * l.space^0
```

syslog/modules/lpeg/sendmail.lua
```diff
@@ -0,0 +1,32 @@
...
+-- l.R(" +","-~")^1) * space
+local sep = l.P',' * space
+local pair = l.Cg(key * "=" * value) * sep^-1
```
trink: inconsistent quote style

syslog/modules/lpeg/sendmail.lua
giganteous: I've ditched the spaces. I'm sorry for the bad copy paste behavior.

syslog/modules/lpeg/sendmail.lua
giganteous: fixed the quoting. It seems you're preferring double quotes.

syslog/modules/lpeg/sendmail.lua
```diff
@@ -0,0 +1,30 @@
...
+
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+local sep = l.P", "
+local key = l.C(l.alpha^1)
+local value = l.C((l.P(1) - sep)^1)
+local pair = l.Cg(key * "=" * value) * sep^-1
+
+local sendmailid = l.Cg(l.alnum^-8 * l.digit^6, "sendmailid") * l.P": "
```
trink: I believe this is too restrictive e.g. ` e6FFrpW22493:` Since it doesn't have a defined format or length:
- leave it alnum like it was and require at least one pair (as I haven't see an error lines with that format format)
- just check that it ends in a digit followed by a colon

trink: you are correct (ignore the comment above): https://fossies.org/dox/sendmail.8.15.2/queue_8c_source.html#l05381

PR #446, opened 2019-08-09
Create client in process_message

bbinet:
this is to avoid socket.connect blocking the creation of the sandbox.

We had a case were the heka_tcp server was listening but the socket.connect would not complete (and it does not seem to timedout neither), so hindsight seems to be blocked creating this output sandbox and the hindsight.cp checkpoint file was not created.

### DIFF
--- a/socket/CMakeLists.txt.socket
+++ b/socket/CMakeLists.txt.socket
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
-project(socket VERSION 3.0.11 LANGUAGES C)
+project(socket VERSION 3.0.12 LANGUAGES C)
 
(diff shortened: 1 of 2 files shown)


trink:
We should be able to set the timeout on the connection. e.g. socket.settimeout(timeout) before the connect call.  As we don't want process_message to block indefinitely either.  Please bump the package version https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/socket/CMakeLists.txt.socket#L6

bbinet:
@trink socket lib does not seem to have a `socket.settimeout(timeout)` method, but I should be able to use a master object instead:
http://w3.impa.br/~diego/software/luasocket/tcp.html#socket.tcp
And set the timeout on this master object before the connect call:
http://w3.impa.br/~diego/software/luasocket/tcp.html#settimeout

I will update the pull request in that direction.

bbinet:
@trink implemented in 7f273f3

trink:
Thanks, sorry for the delay.

PR #455, opened 2019-09-06
Untrust the stat= key; value can contain garbage

giganteous:
I ran into a mail server reply with "Hostname=...." stuff which overwrote the Hostname key set by the parent (syslog) decoder.

### DIFF
--- a/syslog/modules/lpeg/sendmail.lua
+++ b/syslog/modules/lpeg/sendmail.lua
@@ -21,7 +21,9 @@ setfenv(1, M) -- Remove external access to contain everything in the module
 local sep = l.P", "
 local key = l.C(l.alpha^1)
 local value = l.C((l.P(1) - sep)^1)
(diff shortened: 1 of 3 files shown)


syslog/modules/lpeg/sendmail.lua
```diff
@@ -21,7 +21,9 @@ setfenv(1, M) -- Remove external access to contain everything in the module
 local sep = l.P", "
 local key = l.C(l.alpha^1)
 local value = l.C((l.P(1) - sep)^1)
-local pair = l.Cg(key * "=" * value) * sep^-1
+local statpair = l.Cg(l.C(l.P'stat') * l.P'=' * l.C(l.P(1)^0))
```
trink: inconsistent quote style ;)
also `l.P'='` can just be `"="`

giganteous:
My sincere apologies for the delay.

PR #466, opened 2019-10-04
Bug 1585673 - Process the exceptions out of the runs[] history

trink:


taskcluster/io_modules/decoders/taskcluster/live_backing_log.lua
```diff
@@ -882,6 +876,30 @@ function decode(data, dh, mutable)
         end
     end
 
+    -- handle any exception runs reported in the history as they don't have their own resolution messages
+    -- https://bugzilla.mozilla.org/show_bug.cgi?id=1585673
+    for i=0, pj.runId - 1 do
+        local prun = pj.status.runs[i + 1]
+        if prun.state == "exception" and prun.resolved then
```
srfraser: I wonder if this also happens for the `failed` state. If it doesn't then this is good.

trink: https://bugzilla.mozilla.org/show_bug.cgi?id=1585673#c2

PR #469, opened 2019-10-10
Added derived_kind_costs scheduled query, for cost analysis of all ta…

srfraser:
…sk kinds.

### DIFF
--- a/taskcluster/sql/derived_task_costs.sql
+++ b/taskcluster/sql/derived_task_costs.sql
@@ -1,0 +1,58 @@
+WITH
+  a AS (
+  SELECT
+    date,
+    project,
+    platform,
+    kind,
+    workerType,
+    owner,
+    COUNT(*) AS tasks,
+    ROUND(SUM(TIMESTAMP_DIFF(resolved, started, millisecond)) / 3600000, 2) AS hours,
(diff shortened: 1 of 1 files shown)


trink:
This is just a subset of https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/taskcluster/sql/derived_workertype_costs.sql.  Is there a reason that table cannot be used? I thinke we should migrate the test suite costs to use it also and then dump these two tables.

srfraser:
> This is just a subset of https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/taskcluster/sql/derived_workertype_costs.sql. Is there a reason that table cannot be used? I think we should migrate the test suite costs to use it also and then dump these two tables.

It was based on `derived_test_suite_costs`, since I wanted to include more than just tests. It seems we've duplicated some effort, since i didn't see the workertype costs query - I wasn't wanting to query based on worker type, so it didn't make sense to look for that name. 

Assuming it's ok to have a table with large numbers of nulls in, then sure, I can change to the 'workertype' table. I think if I'd caught it earlier I'd suggest a less specific name!

trink:
It is easy enough to rename and alter the schema if necessary at this point.

trink:
Merging to capture the existing state

PR #477, opened 2019-10-31
Updated queries and gcp automation

srfraser:
Updated some of the derivation queries with the needed values. I'll go through and have a clear-up soon, to avoid duplication.

I've added some automation to export the tables required for `mach try fuzzy --show-estimates` using cloud scheduler and a cloud function. Storing those here.

### DIFF
--- a/taskcluster/sql/machtry_quantiles.sql
+++ b/taskcluster/sql/machtry_quantiles.sql
@@ -1,0 +1,18 @@
+WITH
+  a AS (
+  SELECT
+    APPROX_QUANTILES(total_seconds, 100 IGNORE NULLS) AS quants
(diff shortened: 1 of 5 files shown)


taskcluster/sql/machtry_quantiles.sql
```diff
@@ -0,0 +1,18 @@
+WITH
+  a AS (
+  SELECT
+    APPROX_QUANTILES(total_seconds, 100 IGNORE NULLS) AS quants
```
trink: Do we want to take different runs into account above P59 is somewhat bogus as they are greater than a day.  For now merging as is since this is capturing what is already running.

PR #512, opened 2020-04-09
[taskclusteretl] Discover workerTypes with missing costs

srfraser:


trink:
This query could be added as a data source directly to the DataStudio report; if there are other consumers then making it a scheduled query/derived table is fine.  Let me know, r+ on the query

srfraser:
> This query could be added as a data source directly to the DataStudio report; if there are other consumers then making it a scheduled query/derived table is fine. Let me know, r+ on the query

Thanks! I think I'd like in vcs somewhere even as a query data source. Perhaps a new subdirectory here, if it'll cause confusion to have them all in one?

trink:
I wish we could version control the entire report and datasource configuration.  A sub directory or just changing the prefix from derived to datasource is fine.

srfraser:
> I wish we could version control the entire report and datasource configuration.

+1 to that, if that was possible we could also programmatically generate new views.

> A sub directory or just changing the prefix from derived to datasource is fine.

I'll move the files around and update the PR. Thanks!

PR #518, opened 2020-06-29
Elastic tls

giganteous:
I have pulled in the original #201, but I don't think I can push to that branch. I have based my work a bit on it.

I wonder if the module cannot go back to use the plain 'http.request': has support for passing it a connection factory: http://w3.impa.br/~diego/software/luasocket/http.html . Let me know what you think, as it is slightly more work.

I haven't built a test! Perhaps you have a suggestion on if, and how to test this?


trink:
Generally we run tests for I/O plugins like this using their own unique cmake TEST_CONFIGURATION and manually run them in the proper environment (see the aws/gcp tests).  I don't even have an ElasticSearch test environment these days.  I was hoping to do a sanity check code review after you confirmed it works ;) but at some point I will have to setup a test environment if are supporting it.

giganteous:
I've tested this on our TLS and basic auth protected instances, and it is humming nicely.

PR #521, opened 2020-07-09
Issue 199 - lfs tail, reset the checkpoint to 0 on a file open error

trink:


giganteous:
This seems to fix the issue I'm seeing when I rotate often.

I'm testing it on my servers, but since I'm away for a few weeks, you'll have to do without those results for some time.

trink:
Comments from IRC

I have a fix for the off by one cp error you saw too (I will add it to the pr)... the fh:lines() returns the next line with the \n delimiter stripped but it also returns a partial line up to EOF and no delimiter (messing up the manual byte count, so I switched it to seek())

The best fix would be to make a version of lines() that includes the new line so the byte count and eof termination can be easily detected (it would also allow the partial line issue in the tail.lua documentation to be addressed)

PR #525, opened 2020-07-28
lfs - rewrite tail.lua to avoid the io.lines() issues

trink:


lfs/CMakeLists.txt.lfs
```diff
@@ -37,7 +37,7 @@ if(LUA51) # build against installed lua 5.1
     find_program(LUA NAMES lua lua.bat)
     add_test(NAME test COMMAND ${LUA} test.lua)
 else() # lua_sandbox build
-    set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, luasandbox (>= 1.0.2), luasandbox-lpeg (>= 1.0.10)")
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, luasandbox (>= 1.0.2), luasandbox-lpeg (>= 1.0.10), ${PACKAGE_PREFIX}-gzfile (>= 0.0.3)")
```
deric, line 40: `PACKAGE_PREFIX` variable doesn't seem to be set in this context.
Maybe add:
```
set(PACKAGE_PREFIX luasandbox)
```

lfs/sandboxes/heka/input/tail.lua
```diff
@@ -128,84 +129,92 @@ if delimiter then
...
-        fh, checkpoint = open_file(checkpoint)
-        if not fh then return 0 end
+        local t = type(checkpoint)
+        if t == "string" then
+            local i, o = t:match("(%d+):(%d+)")
+            inode  = tonumber(i) or 0
+            offset = tonumber(o) or 0
+        elseif t == "number" then  -- migrate the old checkpoint format
+            inode  = get_inode() or 0
+            offset = t
```
deric, line 204: This probably should be:
```
offset = inode
```
Otherwise `offset` will be assigned string value `"number"`.

trink: it needs to be set to checkpoint... fixed

lfs/CMakeLists.txt.lfs
trink: sigh I just copied this from another file... replaced it with luasandbox

PR #527, opened 2020-08-10
Remove default type as Elastic 7 deprecates document types

deric:
In order to be compatible with recent Elastic versions document type information should not be defined or `_doc` type could be used.

`_doc` is a dummy document type that will work as well with older Elastic versions

See: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/removal-of-types.html

### DIFF
--- a/elasticsearch/CMakeLists.txt
+++ b/elasticsearch/CMakeLists.txt
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 cmake_minimum_required(VERSION 3.0)
-project(elasticsearch VERSION 1.0.8 LANGUAGES C)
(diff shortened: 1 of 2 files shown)


trink:
Changing the defaults is a breaking change for anyone using pre v7.  Ideally the encoder would accept a version and then apply the correct defaults (the version should default to 5 (when nil) since that is what this encoder was originally created against).

Also the elasticsearch package version needs to be incremented 1.0.9
https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/elasticsearch/CMakeLists.txt#L6

deric:
@trink You're right, it would be a breaking change for anyone relying on default values. I wasn't sure how to get the Elastic version, alternatively we could send `GET elasticsearch:9200` and parse the version... but then the type would be determined upon connection to Elastic.

PR #529, opened 2020-09-08
taskcluster: work around changed timestamp format

whd:


whd:
trink graciously looked at this and from his review I've simplified it from the original fix and also fixed the grammar a bit. I'm going to merge this into dev to test its efficacy on the dev cluster.

PR #534, opened 2021-06-19
Update README.md

dai-yamashita:
CMAKE build path is incorrect.


dai-yamashita:
sorry .. PATH was there.

